PR #11, opened 2017-07-17
Fix compilation in GCC 4.6.4

nattthebear:
The fixed code had undefined behavior due to multiple modifications to the same variable within a single sequence point.


LIJI32:
Merged, thanks! :)

Issue #5, opened 2017-04-24
Libretro port runloop issues


twinaphex:
Hi there,

@not6 did most of the work so far on doing a basic libretro port. We have made some nice progress.

Our only problem so far though is that the runloop that has to be implemented doesn't really fit our model.

It would be best if we could run exactly one frame once every retro_run call, without needing to have a vblank handler active.

Because the runloop is not implemented right now in a way libretro understands, fastforwarding and rewinding is currently not possible.

You can see what we currently do for SameBoy here -

https://github.com/libretro/SameBoy/blob/master/libretro/libretro.c#L272

This is our vblank callback implementation -

https://github.com/libretro/SameBoy/blob/master/libretro/libretro.c#L142

You can see here how we implemented the runloop for another GBC/GB emulator, Gambatte - 

https://github.com/libretro/gambatte-libretro/blob/master/libgambatte/libretro/libretro.cpp#L829

Thanks for the great work on this emulator.

LIJI32:
Sure, I can add an API for that; just a few questions:
1. I assume this means libretro expects emulator to not handle any FPS capping, so nanosleep should not be called when this API runs, am I right?
2. The vblank interval is not always constant on the Game Boy, due to being able to turn the screen on and off. Assuming it is constant will causes noticeable sound pops and jumps in some games (Pokémon Pinball for example). How would libretro handle that? Should the actual time passed be returned?
3. Any special requests for sounds? Ability to control the buffer size directly?

I'll be going on a ~2 week vacation tomorrow morning and I won't be able to work on SameBoy during that time, hopefully if you answer quick enough I'll be able to commit tonight. :)

twinaphex:
Hi there,

1. FPS capping is not needed for a libretro port, what we do is we set the audio rate and the refresh rate on the libretro core side and then the FPS capping is done by the frontend itself. So the core should stay mostly out of the way of such things. So nanosleeping and the like is all unnecessary and is best left avoided.
2. Not sure about this or how Gambatte deals with it.
3. No special requests really. The best way would be is if we dont' have to set a sound callback and we can just grab a 16-bit sound buffer for each retro_run iteration with a 'size' variable and then just pass that off to the audio_batch_cb callback for audio output.

Have a good vacation!

LIJI32:
1. Assumed so, no problem. I'll add the API in a few minutes.
[...]

PR #14, opened 2017-10-09
libretro core

fr500:
@LIJI32 in case you're interested, the core is more or less complete, I plan to clean it up a bit more in the next few days but it's generally usable now.

If there are any changes needed for upstreaming make sure to let me know.


fr500:
I can also squash it in a single commit if you prefer.

LIJI32:
It generally looks good, although I'm not entirely happy with the way the __LIBRETRO__ and HAVE_DEBUGGER defines are used in the Core tree, as well as how the precompiled Boot ROM is hardcoded instead of auto-generated, but I will probably fix those myself tomorrow, so expect a merge then. :)

While we're at it, I have a few questions about the core since my experience with libretro is not very extensive:

1. Which platforms libretro supports are compatible with the SameBoy core? And which aren't, and why? If there are any changes required on my end, they should probably be very easy to do and will allow to core to run on even more platforms.
2. Are there any libretro features the other Game Boy cores support and SameBoy doesn't? What has to be done to support them?

fr500:
Hey.

1. All of them should work I guess the problem is catering to the different build systems of a specific platform. For android I still have to figure out how to make the jni makefile and it should just work.

That said, I find the core too slow to be used on consoles / web though, but maybe it's something wrong on our end, gambatte runs at like 3000fps, sameboy barely goes over 60fps on fast forward. Maybe we're still being limited somewhere.

2. Not really, actually since you support the gameboy camera it could be the very first core that supports the libretro camera API properly. Not really sure about which changes are required for this to happen.

One killer feature would be being able to run two instances of sameboy as a single core to implement link-cable, but I guess for that I have to wait for you to have working link cable support. In gambatte we have a very rudimentary link cable over TCP implementation but it's not really usable.

Another thing that is not hooked up is the memory API that allows us to support RetroAchievements. I have never hooked that up but maybe @leiradel can give us a hand there.

https://github.com/libretro/gambatte-libretro/blob/master/libgambatte/libretro/libretro.cpp#L747

I saw there is a memory_direct function I may be able to use in SameBoy to hook this up.

I'm not all that familiar with the gameboy hardware but tonight I'll test upstream to see if there are any features I'm missing.

LIJI32:
1. Awesome. SameBoy *is* indeed slower than Gambatte, but 60FPS on fast forward is definitely something gone wrong, so I'll have to check it. If nothing goes wrong, version 0.11 is expected to get a major speed up.
2. Awesome as well! Actually, there IS already link-cable/serial API, and it's already used for Game Boy Printer emulation (which can be used with the Cocoa frontend). Not sure what else libretro requires to get it to work over TCP, but judging by the recent video about libretro and netplay it theoretically shouldn't be hard to get it working. There's also a rumble API (used by games such as Pokémon Pinball) which is should be extremely easy to integrate into the libretro core, and it feels *really* sweet on the WIP iOS port.

fr500:
Well the link cable support wouldn't need anything.
Basically the idea would be to have a single Sameboy libretro core that runs two instances side by side. The rest is handled by using savestates :)

I'll check how rumble works in other cores, I think hooking it up is not all that difficult.

fr500:
Also sound seems very crackly on this core for some reason but I really dunno much about audio.

LIJI32:
Version 0.10 introduced a new APU that renders sounds at 2MHz, and then downsamples it and applies a high-pass filter (similar to hardware). It sounds REALLY sweet on both the Cocoa and the SDL port on my machine, so it's probably something done wrong in the integration (probably related to differences in how SameBoy and libretro sync video and audio).

I attached an audio capture from the Cocoa port (although it should sound the same in SDL) for reference:
[sameboy_audio.flac.zip](https://github.com/LIJI32/SameBoy/files/1373766/sameboy_audio.flac.zip)

fr500:
Ah it seems this has been discussed before :p
https://github.com/LIJI32/SameBoy/issues/5

Did you implement a way to query how many samples are in the buffer?
I assume it's this:

size_t GB_apu_get_current_buffer_length(GB_gameboy_t *gb);

fr500:
Seems my latest commit improved audio but I still hear some crackles
Tried to implement what @aliaspider mentioned in that issue I linked above.

Not sure if I did it correctly though :p

Awakened0:
The audio's a lot better after that commit. Now I only notice slight pops here and there and some "grainyness" in high pitched notes, like Ballad of the Wind Fish in LA DX, when compared to Gambatte. There are also some nasty pops whenever emulation is paused or if the shader is changed on the fly.

An option I'd like to have from Gambatte is "Color Correction". With it off it looks like this:

legend of zelda the - link s awakening dx usa europe sgb enhanced restoration by artemis251 v1 1 -171011-010237

And with it on:

legend of zelda the - link s awakening dx usa europe sgb enhanced restoration by artemis251 v1 1 -171011-010412

LIJI32:
The Game Boy APU has a DC offset which causes the pops when pausing the emulator, on hardware the DC offset is removed by an analog high-filter while on SameBoy calling GB_set_highpass_filter_mode with either GB_HIGHPASS_ACCURATE or GB_HIGHPASS_REMOVE_DC_OFFSET (The high-pass filter is off by default). I'll try to see what happens in Ballad of the Wind Fish, but increasing the sample rate from 44.1KHz to 96KHz is likely to improve this as well. As for color correction, it's probably simple enough for me to add it to v0.10.

leiradel:
As long as there's an API for getting pointers to and sizes of the emulated memory, I can add Retro Achievements support. I'll take a look.

LIJI32:
This is exactly what GB_get_direct_access does :)

fr500:
ah that's what the DC OFFSET does, I was actually testing with that enabled but pushed with that disabled LOL.

I'll make it a core option.

leiradel:
I'm having trouble mapping the GB_DIRECT_ACCESS values to the actual memory I need to access:

* Zero Page: ?
* RAM Bank 0: ?
* RAM Bank 1 (for CGB): ?
* Save RAM: GB_DIRECT_ACCESS_CART_RAM
* VRAM: GB_DIRECT_ACCESS_VRAM
* OAM: GB_DIRECT_ACCESS_OAM

In Gambatte I also map ROM banks 0 and 1, but those (and VRAM actually) aren't necessary for achievements.

LIJI32:
Assuming zero-page refers to High RAM, it's GB_DIRECT_ACCESS_HRAM. GB_DIRECT_ACCESS_RAM gives you access to all RAM banks (1 bank on DMG, 8 banks on CGB; they're mapped consecutively on memory). ROM behaves the same way using GB_DIRECT_ACCESS_ROM.

LIJI32:
Added color correction to the core with 4 modes; can be configured with GB_set_color_correction_mode:

 * `GB_COLOR_CORRECTION_DISABLED` - No color correction is made at all, use direct RGB translation (Same as v0.9)
 * `GB_COLOR_CORRECTION_CORRECT_CURVES` - Perform non-linear RGB translation which slightly increases contrast
 * `GB_COLOR_CORRECTION_EMULATE_HARDWARE` - Same as previous, but additionally "ties" the value of the green channel to the blue channel, which causes some hue shifting
 * `GB_COLOR_CORRECTION_PRESERVE_BRIGHTNESS` - Same as previous, but also applies brightness preservation (more accurately, the "locks" the Luminance value in the HSL model), as the previous correction darkens green-yellow colors and lightens purple-magenta hues.

image

LIJI32:
Merged into a side-branch, will be merged into master once I clean it up a bit

Awakened0:
Nice, I like the "Emulate Hardware" one the most :)

leiradel:
I can register achievements for Super Mario Land with SameBoy. For example, [Fire Power](http://retroachievements.org/Achievement/1144) should be awarded by the middle of the first stage. It's expression is `0xHffb5=2_0xHffff=7`.

This achievement, and most if not all of the others, test the memory at 0xFFFF, which strikes me as odd, since this is the IE register for GB/GBC.

@LIJI32 do you write back the value of this register to memory, so that accessing the value via the pointer obtained from `GB_get_direct_access` with `GB_DIRECT_ACCESS_HRAM` reads the correct value?

LIJI32:
IE is an oddball – it is a memory mapped register (similar to IF and LCDC) but it is not located with the rest of the memory mapped registers (FF00-FF7F). Instead, it is located right after the high-ram area (which is just plain-old RAM, located in the range FF80-FFFE). For that reason, IE can't be accessed using `GB_get_direct_access` as it is technically is neither sections. The "official" way to access it is using GB_memory_read(gb, 0xffff), and the "dirty" way (requires GB_INTERNAL) is gb->interrupt_enable.

leiradel:
Unfortunately neither will work with achievements. IE is accessed as if it was a regular memory location right after the end of the zero page (high RAM), so I need the pointer returned by `GB_get_direct_access` to have its value copied at offset 0x7F.

If you provide a `GB_DIRECT_ACCESS_IE` which returns a pointer with size=1 I believe I can make it work, or just extend `GB_DIRECT_ACCESS_HRAM` to include IE.

LIJI32:
A bit hacky, but considering the fact IE is the only memory address inaccessible via direct_access, adding `GB_DIRECT_ACCESS_IE` seems reasonable. I'll add it in a minute. :)

LIJI32:
I cleaned up the libretro_core branch a bit. The major changes you should be aware of are:
 * .o files are now saved in build/obj (to prevent clutter) and have a _libretro suffix (to prevent conflicts with other targets, as they use other flags). The suffix also allowed renaming gbmemory back to memory.
 * The "baked" Boot ROMs are now auto-generated, so they don't need to manually be created when I update the source asm files. This make rgbds and xxd build dependencies.
 * The final output file is now at build/bin/sameboy_libretro.*
 * The libretro-specific Makefiles were moved to libretro's folder. The core can be built from the folder or from the root folder using `make libretro`

I only tested these changes on my machine (macOS) and not any of the other platforms libretro supports (I don't have that many toolchains installed :P), so please verify I didn't break anything and I'll merge the branch into master.

Alcaro:
> Unfortunately neither will work with achievements.

You could add your own memory map for that byte and read it from GB_memory_read every frame.

But yeah, GB_DIRECT_ACCESS_IE is less hacky.

> The suffix also allowed renaming gbmemory back to memory.

Renaming memory.h is only relevant if that one would be found by a #include from the system headers. As far as I know, this this is only the case if there's a `-I Core` (or similar) nearby, which I couldn't find.

Either I'm wrong and it's always harmful (maybe some Android-internal makefile sets a few extra -I), in which case renaming that file back won't help, or renaming it was never needed. Hopefully it's the latter.

> This make rgbds and xxd build dependencies.

We kinda don't like build dependencies.

But yes, not adding extra steps is also important.

I propose this compromise: Keep that build rule around, but keep the output on github as well, and don't delete it on `make clean` (or only delete if rgbasm exists - not sure if that's worth it, or even possible).

That way, you don't need to do anything manually (unless you `git add` specific files rather than `git commit -a`, or do edit-commit-compile rather than edit-compile-commit, or something like that), without forcing everyone to install rgbasm.

> make -C libretro

Better use `$(MAKE) -C libretro`, so it'll pass on command line arguments (like `make platform=osx` or `make -j8`) to the child.

leiradel:
> You could add your own memory map for that byte and read it from GB_memory_read every frame.

The pointers are saved once and reused every frame, so I can't read IE every frame, the value must already be available via some pointer to be read.

> But yeah, GB_DIRECT_ACCESS_IE is less hacky.

I agree.

LIJI32:
> Renaming memory.h is only relevant if that one would be found by a #include from the system headers. As far as I know, this this is only the case if there's a -I Core (or similar) nearby, which I couldn't find.

Oh, so *THAT* was the problem, the commit didn't specify .c or .h and for some reason I assumed it was the .c file. So you're right, there's an extraneous -I Core option (it also affects the non-libretro frontend). I'll get rid of it. :)

> Keep that build rule around, but keep the output on github as well

Kind of beats the purpose of that change, as it keeps generated files committed to the repository, and I'd still have to remember to update this file whenever I update the boot ROMs. (The build system DOES support, however, specifying BOOTROMS_DIR for a folder with precompiled boot ROMs, which can be used if you don't have rgbasm)

> Better use `$(MAKE) -C libretro`
Good point, I'll fix that.

fr500:
The only problem I have with this is that the libretro buildbot expects the build product in the current working dir.

I understand you made this to keep things tidier but it's a pita to address...

fr500:
Oh also the autogenerated boot roms, I don't think those tools are even available on MSYS2 are they?

leiradel:
> The pointers are saved once and reused every frame, so I can't read IE every frame, the value must already be available via some pointer to be read.

Just to clarify: I could call `retro_get_memory_data/size` every frame and deal with IE there, but that won't work when using the memory map interface, which is not only the preferred way in the achievements code but also the only way to deal with the zero page RAM from GB/GBC.

LIJI32:
> buildbot expects the build product in the current working dir

How awful it is to work around this? In the worst case, we could have the "official" and libretro repositories differ in the location of these files, it's a minor enough change to not cause any conflicts or prevent mutual merges.

> I don't think those tools are even available on MSYS2

rgbds has a Windows port (Which can be used in MSYS2), xxd can be replaced with an equivalent hexdump line (Which is included with MSYS2)

leiradel:
@Alcaro do you think those values are wrong?

```
RetroArch [INFO] :: Environ SET_MEMORY_MAPS.
RetroArch [INFO] ::    ndx flags  ptr              offset   start    select   disconn  len      addrspace
RetroArch [INFO] ::    001 M1A1bc 000000006CF67B80 00000000 0000FF80 00000000 00000000 0000007F
RetroArch [INFO] ::    002 M1A1bc 000000006CF5FB75 00000000 0000FFFF 00000000 00000000 00000001
RetroArch [INFO] ::    003 M1A1bc 0000000003E721B0 00000000 0000C000 00000000 00000000 00002000
RetroArch [INFO] ::    004 M1A1bc 0000000000000000 00000000 0000A000 00000000 00000000 00002000
RetroArch [INFO] ::    005 M1A1bc 0000000003E62D40 00000000 00008000 00000000 00000000 00002000
RetroArch [INFO] ::    006 M1A1bc 000000006CF67D25 00000000 0000FE00 00000000 00000000 000000A0
RetroArch [INFO] ::    007 M1A1bC 00000000043D6890 00000000 00000000 00000000 00000000 00004000
```

These are the descriptors reported by gambatte:

```
RetroArch [INFO] :: Environ SET_MEMORY_MAPS.
RetroArch [INFO] ::    ndx flags  ptr              offset   start    select   disconn  len      addrspace
RetroArch [INFO] ::    001 M1A1bc 0000000003C238D8 00000000 0000FF80 FFFFFF80 00000000 00000080
RetroArch [INFO] ::    002 M1A1bc 00000000042CC4B0 00000000 0000C000 FFFFF000 00000000 00001000
RetroArch [INFO] ::    003 M1A1bc 00000000042CD4B0 00000000 0000D000 FFFFF000 00000000 00001000
RetroArch [INFO] ::    004 M1A1bc 0000000000000000 00000000 0000A000 FFFFE000 00000000 00002000
RetroArch [INFO] ::    005 M1A1bc 00000000042C84B0 00000000 00008000 FFFFE000 00000000 00002000
RetroArch [INFO] ::    006 M1A1bc 0000000003C23758 00000000 0000FE00 00000000 00000000 000000A0
RetroArch [INFO] ::    007 M1A1bC 00000000042B84B0 00000000 00000000 00000000 00000000 00004000
```

Alcaro:
I don't remember the exact rules for how memory maps work, but it doesn't look obviously wrong.

leiradel:
I find it odd that all select values for the SameBoy descriptors are zero. The code that creates the descriptors is this:

```C++
   struct retro_memory_descriptor descs[7];
   size_t size;
   uint16_t bank;

   memset(descs, 0, sizeof(descs));

   descs[0].ptr   = GB_get_direct_access(&gb, GB_DIRECT_ACCESS_HRAM, &size, &bank);
   descs[0].start = 0xFF80;
   descs[0].len   = 0x007F;
   
   descs[1].ptr   = GB_get_direct_access(&gb, GB_DIRECT_ACCESS_IE, &size, &bank);
   descs[1].start = 0xFFFF;
   descs[1].len   = 1;

   descs[2].ptr   = GB_get_direct_access(&gb, GB_DIRECT_ACCESS_RAM, &size, &bank);
   descs[2].start = 0xC000;
   descs[2].len   = 0x2000;
      
   descs[3].ptr   = GB_get_direct_access(&gb, GB_DIRECT_ACCESS_CART_RAM, &size, &bank);
   descs[3].start = 0xA000;
   descs[3].len   = 0x2000;
      
   descs[4].ptr   = GB_get_direct_access(&gb, GB_DIRECT_ACCESS_VRAM, &size, &bank);
   descs[4].start = 0x8000;
   descs[4].len   = 0x2000;
      
   descs[5].ptr   = GB_get_direct_access(&gb, GB_DIRECT_ACCESS_OAM, &size, &bank);
   descs[5].start = 0xFE00;
   descs[5].len   = 0x00A0;
      
   descs[6].ptr   = GB_get_direct_access(&gb, GB_DIRECT_ACCESS_ROM, &size, &bank);
   descs[6].start = 0x0000;
   descs[6].len   = 0x4000;
   descs[6].flags = RETRO_MEMDESC_CONST;

   struct retro_memory_map mmaps;
   mmaps.descriptors = descs;
   mmaps.num_descriptors = sizeof(descs) / sizeof(descs[0]);
   environ_cb(RETRO_ENVIRONMENT_SET_MEMORY_MAPS, &mmaps);
```

Alcaro:
Some of the fields are optional, and if zero, are calculated from other fields. I believe those two are sufficient.

Though the ROM seems kinda screwy. There's a button somewhere to allow bankswitching, check docs in libretro.h and do that.

Either way, there's only one true way of finding if they work: try it.

leiradel:
> Either way, there's only one true way of finding if they work: try it.

It doesn't, random achievements are awarded for nothing and the correct ones aren't.

> Though the ROM seems kinda screwy. There's a button somewhere to allow bankswitching, check docs in libretro.h and do that.

> I've removed the ROM descriptor and didn't help.

@LIJI32 does anything in the code I've posted above strike you as wrong?

LIJI32:
It looks good to me, but I'm not sure how libretro handles memory banking. (4000-8000 (ROMX) is banked on both models, 8000-A000 (VRAM) and D000-E000 (CGB RAM) are banked on CGB)

leiradel:
For achievements the ROM contents aren't necessary, I'm more worried about RAM. I need RAM banks 0 and 1 accessible in `descs[2]`. Are they contiguous in memory? If not, I can split `descs[2]` in two and deal with each bank separately, as long as there's some way to get the pointers.

LIJI32:
They're only contiguous if CGB RAM bank 1 is mapped (So you'll need to split them). The easiest way to see if that's the problem though is testing achievements on non-CGB games, since DMG does not have banking.

leiradel:
> They're only contiguous if CGB RAM bank 1 is mapped

I mean contiguous when accessed via the pointer returned by `GB_get_direct_access`.

> The easiest way to see if that's the problem though is testing achievements on non-CGB games, since DMG does not have banking.

I'm testing Super Mario Land for DMG.

LIJI32:
> I mean contiguous when accessed via the pointer returned by GB_get_direct_access.

The pointer `GB_get_direct_access` returns is for the "static" RAM at C000, followed by RAM banks 1-7, so yes, it's contiguous as well.

> I'm testing Super Mario Land for DMG.

So seems like RAM banking isn't the problem then. =/

leiradel:
@Alcaro the cheevos code is mapping all addresses to the first memory descriptor, high RAM, at offset zero. Do you mind taking a look? All those bit operations are way over my head...

```C++
      // This is in cheevos.c, function cheevos_parse_guest_addr
      desc = system->mmaps.descriptors;
      end  = desc + system->mmaps.num_descriptors;

      for (; desc < end; desc++)
      {
         if (((desc->core.start ^ var->value) & desc->core.select) == 0)
         {
            unsigned addr = var->value;
            var->bank_id  = (int)(desc - system->mmaps.descriptors);
            var->value    = (unsigned)cheevos_reduce(
               (var->value - desc->core.start) & desc->disconnect_mask,
               desc->core.disconnect);

   		if (var->value >= desc->core.len)
               var->value -= cheevos_highest_bit(var->value);

            var->value += desc->core.offset;
            RARCH_LOG("[CHEEVOS]: Address %X set to descriptor %d at offset %X\n", addr, var->bank_id + 1, var->value);
            break;
         }
      }
```

This code maps the address to the correct places when using Gambatte. The memory descriptors are the same as posted above.

Alcaro:
I guess it wants select set. Those are the bits that must match the mapping to be relevant; for C000-CFFF, that is 0xF000.

I thought it autogenerated select from length if not set, but maybe not.

leiradel:
> I thought it autogenerated select from length if not set, but maybe not.

It does, In the Gambatte core, all selects are zero with the exception of the save RAM, which is initialized with `(size_t)~0x1FFF`.

I'll try to set the selects by hand and see if cheevos start working with SameBoy.

leiradel:
I've managed to make it work by setting the length of the zero page to 128 instead of 127. The automatic generation of the select field was getting confused by the odd length.

Since I've fork the repository from libretro, I'll wait until @fr500 pulls the commit with the addition of GB_DIRECT_ACCESS_IE.

Thanks everyone, that's one more GB/GBC core with achievements support!

fr500:
Gonna try now
but it's gonna break our buildbot :p

fr500:
@LIJI32 are you on any IRC or discord channel? need to ask you a few questions.

fr500:
Anyway, it builds, the outpu dir is a problem for the buildbot though.
Also the build is full of errors for some reason... at least on windows, and it keeps using clang for some files...

https://hastebin.com/huqaleloti.swift

It even produced a valid binary without clang installed.

fr500:
Hooked up the color filter and the highpass filter

LIJI32:
> @LIJI32 are you on any IRC or discord channel? need to ask you a few questions.

No, but I can join one if it helps

> Also the build is full of errors for some reason... at least on windows, and it keeps using clang for some files...

Yeah, my bad, it's generating .dep files for the .C files when it's assembling the boot ROM files. I'll fix it later today.

leiradel:
I've managed to build the core for Windows with MSYS2 with this command:

```
make -j 4 -f Makefile.libretro platform=win
```

leiradel:
Hm, talked too soon. I'm not able to build after pulling latest.

fr500:
build with 
`make -j4 libretro `

It's generating some errors but it's building just fine

fr500:
@LIJI32 is it ok if we keep prebuilt bootroms in our fork?

LIJI32:
> is it ok if we keep prebuilt bootroms in our fork?

Sure, just make sure you rebuild them whenever you pull. (There's a pending commit to the CGB boot ROM in a side branch, it's going to change before the final 0.10 release)

LIJI32:
My latest commit to the libretro_core branch solved dependency generation which caused the errors on Windows

fr500:
Thanks
There is still one oddity.

We have a frontend feature that checks SRAM contents every T interval, and if there are changes to such contents it autosaves, that's to prevent data loss on an eventual crash or when the android garbage collector decides to kill the unfocused APP.

It is working well with sameboy, actually too well, no matter what happens it always detects changes on every interval.

Is it possibly because of the RTC? I've been testing with link's adventure DX.

LIJI32:
Which buffers are you comparing? If you compare the contents of the .sav file then yes, it's the RTC – as it stores the current in-game time as well as the system time (as a Unix timestamp).

fr500:
We don't use the file based functions at all, but yes it includes the RTC
https://github.com/libretro/SameBoy/blob/master/libretro/libretro.c#L456

Guess it makes sense, maybe, actually RA shouldn't flush the file if only the RTC changed....

Awakened0:
I still get little sound pops on the libretro port even with highpass set to accurate or remove dc offset. It's pretty noticeable on the save select screen or even the main Zelda theme on the beach in LA DX. Gambatte and mGBA are much smoother sounding, though they sound more filtered.

Setting the highpass to accurate or remove dc offset fixes the pop when emulation is paused. However, @fr500 , it seems like it's not applying that option at startup. When I restarted RA I was getting the emulation pause pop until I went into core options and changed high pass to off and back to accurate. Maybe that reinitialized it.

fr500:
Ah! I'll take a look thanks!

LIJI32:
I hear some soft pops in the save select screen in both the Cocoa port and in RetroArch, while I do not hear them on my CGB. Assuming no emulation bugs (The newly written APU was heavily tested and is sample-accurate for almost all cases), and considering the fact the pop have very consistent timing (just before some specific notes start, so it's not a "driver" bug), I believe the reason the sound pops on SameBoy and not on a CGB is the lack of a low-pass filter. I'll test if the sound pop is any more noticeable on my SGB2.

However, I also heard some more noticeable pops only using the RetroArch frontend, so it's definitely something we're doing wrong with the integration. Sadly I don't know much about how libretro handles audio, but I'll trying giving it a look in the weekend.

fr500:
This is your discord if you want to communicate directly and in realtime
https://discord.gg/WzTGy5P 

And we're on #retroarch at freenode.net if you prefer IRC

fr500:
hooked up really strong rumble :p

Awakened0:
Something went wrong. I updated the core from the buildbot and now regular GB games just output a black screen with sound :{ 

Color games still work.

fr500:
Maybe something's up with the prebuilt bios images

fr500:
hmmm no, doesn't seem like it's the case...
weird stuff.

fr500:
seems setting color correction on DMG causes the problem, I'll fix it

fr500:
Fixed in https://github.com/libretro/SameBoy/commit/55412d709367789ba85f33d661e3779c89a8a80c

hyarsan:
The 2xBR video filter just shows a blank screen. I really like 2xBR though...

PR #22, opened 2018-01-02
libretro: Fix video being broken with Vulkan

Nadia-h:
This corrects the pitch parameter for the video refresh callback to be the actual pitch instead of 0.

Setting it to 0 seems to work fine for OpenGL, but it causes the graphics to be broken with Vulkan, at least with an AMD card on Linux.


hyarsan:
This also fixed the 2xBR filter as a side effect!

PR #27, opened 2018-01-27
Update libretro

fr500:
Just a heads up so you can cherry pick changes, the only important changes for you are the memory changes and the pitch changes in the libretro core from retrowertz.


LIJI32:
Done. Also, did the Switch port really didn't require any code changes?

fr500:
Nope, that's the cool thing about libretro :)
Most of the time it's just makefile stuff

LIJI32:
I know the libretro part would just work, I'm more surprised about the actual core code "just working" with libtransistor's headers and objects :P

PR #45, opened 2018-03-14
Fix some broken shaders 

LMLB:
This fixes Bilinear, SmoothBilinear, and AAOmniScaleLegacy so they aren't sampling at texel edges (which causes one or the other to be chosen semi-randomly, depending on rounding errors).
This also fixes issues with specific scale factors in OmniScaleLegacy and SmoothBilinear.


LIJI32:
Thanks a lot for that! I wanted to fix that for quite a while but wasn't sure what was the correct way to do it. :)

Issue #17, opened 2017-10-29
Can't open ROMs while VisualBoyAdvance is installed


noiob:
(Using macOS High Sierra)

When trying to open any .gbc file, I ran into the following error:
<img width="532" alt="bildschirmfoto 2017-10-30 um 00 20 08" src="https://user-images.githubusercontent.com/8197071/32149853-23434b56-bd0b-11e7-9d4a-84a58be2e01e.png">

The problem was apparently caused by VisualBoyAdvance, since removing it from my system made everything work fine. It might be some kind of conflict concerning file type association, since "Open with" didn't show SameBoy when VBA was installed, but did afterwards, and the "Open" dialog in the app didn't let me select the .gbc files at all.

LIJI32:
This is a common issue with how macOS handle file associations for filetypes without a standard UTI – an older installed application sometimes "prevent" a newly installed one from opening files with the same extension.

Can you try clearing the launchservices cache with the following command and see if it helps?

```
/System/Library/Frameworks/CoreServices.framework/Frameworks/LaunchServices.framework/Support/lsregister -kill -r -domain local -domain system -domain user
```

noiob:
I've resolved the issue by uninstalling VBA, should I try recreating it and then run that command?

LIJI32:
Well if the problem's solved I guess it's all good. Worth to mention it's neither SameBoy's or VBA's fault, it's a bug with macOS's launchservicesd (I had similar issues with SameBoy and mGBA, which I solved by clearing the cache)

PR #63, opened 2018-04-11
Refactor android jni makefiles (#17)

fr500:
This doesn't include the prebuilt roms patch so it can be upstreamed.


LIJI32:
While we're at it, the CGB and AGB boot ROMs were recently updated, you should rebuild the prebuilt ones.

PR #68, opened 2018-04-19
Fix savestates in SDL2 port

lotharsm:
Because SDL_SCANCODE_0 comes *after* SDL_SCANCODE_9 in the SDL keycode
table, we have to check if the keycode is between >=1 and <=0. We also
have to substract SDL_SCANCODE_1 in order to set command_parameter
properly.

See https://wiki.libsdl.org/SDLScancodeLookup for reference.

Errata: Currently, the savestate created with CTRL+0 is created, but
refuses to load on Windows (working fine on Linux).

### DIFF
--- a/SDL/main.c
+++ b/SDL/main.c
@@ -205,9 +205,9 @@ static void handle_events(GB_gameboy_t *gb)
                         
                     default:
                         /* Save states */
-                        if (event.key.keysym.scancode >= SDL_SCANCODE_0 && event.key.keysym.scancode <= SDL_SCANCODE_9) {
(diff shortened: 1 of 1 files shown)


LIJI32:
What do you mean refuses to load on Windows? Is GB_load_state not being called? Or does it return some sort of error?

lotharsm:
Yes, GB_load_state is not being called. It seems that for some reason the CTRL+0 key combination is not recognised by the Windows port at all. I tested it under Linux and it works fine there. I think it's some sort of oddity of the SDL2 library.

PR #70, opened 2018-04-22
SDL2: Update to SDL_OpenAudioDevice()

lotharsm:
Instead of the legacy SDL_OpenAudio() method, we now use the newer
SDL_OpenAudioDevice() functions. This fixes audio in Windows if the SDL
version is 2.0.6 or higher.

It also allows us to use 48kHz audio for Windows (96kHz somewhat works
too, but since we don't get absolutely smooth audio with it, I'd stick
with 48kHz for now until we find a solution. 44.1Khz is available as
fallback for SDL 2.0.5 and lower.

And finally, thanks to SDL 2.0.6+ offering support for the WASAPI audio driver,
we now can reduce the audio buffer length to 32. Might be a bit too low, so maybe
we need to increase this later on.

Yes, the 2.0.5 to 2.0.6 transition was
quite harsh in terms of Windows audio support...

### DIFF
--- a/SDL/main.c
+++ b/SDL/main.c
@@ -10,8 +10,15 @@
 #ifndef _WIN32
 #define AUDIO_FREQUENCY 96000
 #else
-/* Windows (well, at least my VM) can't handle 96KHz sound well :( */
-#define AUDIO_FREQUENCY 44100
+/* LIJI32 says: Windows (well, at least my VM) can't handle 96KHz sound well :(
+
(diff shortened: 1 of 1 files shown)


rootfather:
Still a few audio dropouts with a buffer size of 32. Funny enough, it seems that decreasing the buffer size improves the audio quality, e.g. 512 samples as buffer size sounds much worse. I think I have to find the sweet spot somehow... Absolutely not satisfied with this non-scientific way of fixing this. :(

PR #95, opened 2018-07-03
mingw libretro

kps501:


LIJI32:
There is already an implementation of `vasprintf` for Windows in `Windows/stdio.h`. Also I don't think `#ifndef vasprintf` would work. Which header is supposed to define it?

kps501:
I don't know. Was playing around with mingw in appveyor image and it kept complaining about it missing. I managed to finally compile my own 32-bit windows libretro dlls by adding that macro.

But I guess it's a bad idea so I'd better trash it. Thanks!

PR #107, opened 2018-08-13
fix undefined reference

fr500:
Rebased the core and ran into this, not sure it's the way you want it fixed but... well it fixed it.
I can fix it properly if you want I just wanted it building again.


LIJI32:
Not ideal but should work, how do other cores use randomly generated values?

PR #113, opened 2018-10-12
Use clang only if available

RobLoach:
Redefining CC isn't a standard thing to do in Makefiles. See more information about this over at:
https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,8 +34,6 @@ OBJ := build/obj
 BOOTROMS_DIR ?= $(BIN)/BootROMs
 
 # Set tools
-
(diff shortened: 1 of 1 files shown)


RobLoach:
Having it switch to `clang` if it's available is a great idea :+1:

RobLoach:
@LIJI32 Would you mind testing out this update? Thanks a lot for your patience on this.

LIJI32:
Seems to work, and is still overridable using `make CC=gcc`

PR #114, opened 2018-10-12
Fix control reaches end of non-void function

RobLoach:
This change makes it so that there is a return value when a kind doens't match. Allows -Werror=return-type to pass.

```
Core/debugger.c: In function ‘read_lvalue’:
Core/debugger.c:239:1: error: control reaches end of non-void function [-Werror=return-type]
 }
```


RobLoach:
The other way to keep it passing is to just return at the end. Both Clang and GCC should be okay with that, right?

LIJI32:
This should work, just have it return `VALUE_16(0)` instead to keep it simpler and shorter. (That line will never run anyway)

PR #115, opened 2018-10-12
Fix multichar character constant warning

RobLoach:
Multi-character character constants are implementation defined, and therefore they are risky to use for use when trying to be compatible with so many different platforms.

```
Core/gb.c: In function ‘GB_is_inited’:
Core/gb.c:423:25: error: multi-character character constant [-Werror=multichar]
     return gb->magic == 'SAME';
                         ^~~~~~
Core/gb.c: In function ‘GB_reset’:
Core/gb.c:560:28: error: multi-character character constant [-Werror=multichar]
     gb->magic = (uintptr_t)'SAME';
                            ^~~~~~
```

This change turns the multi-character into a define with the converted integer.

### DIFF
--- a/Core/gb.c
+++ b/Core/gb.c
@@ -16,6 +16,8 @@
 #define GB_rewind_push(...)
 #endif
 
+#define GB_SAME 1396788549
+
 void GB_attributed_logv(GB_gameboy_t *gb, GB_log_attributes attributes, const char *fmt, va_list args)
 {
     char *string = NULL;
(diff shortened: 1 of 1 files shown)


LIJI32:
I'm well aware they're implementation defined, which is exactly the reason I'm using them in the first place. Since save states are not serialized in a "standard" way the save state format is actually implementation defined. Specifically, save states created on little-endian targets are not compatible with big-endian targets and vice-versa. Previous versions of SameBoy were similarly incompatible between 32- and 64-bit targets, which is why the `magic` field used to be a `uintptr_t`. The correct solution is to disable this warning for affected compilers.

PR #116, opened 2018-10-12
Fix boolean check against integer

RobLoach:
```
Core/z80_cpu.c: In function ‘rlc_r’:
Core/z80_cpu.c:1145:17: error: ‘<<’ in boolean context, did you mean ‘<’ ? [-Werror=int-in-bool-context]
     if (!(value << 1)) {
          ~~~~~~~^~~~~
cc1: all warnings being treated as errors
```


LIJI32:
I'd rather this warning disabled instead, so GCC doesn't break over warnings Clang doesn't issue. Since Clang is the only compiler I use on day-to-day development on SameBoy, any extra warning GCC issues is potentially a build breaker I won't be aware of.

PR #117, opened 2018-10-12
Fix parentheses around ‘+’ in operand of ‘&’

RobLoach:
This fixes the parentheses around the +...

```
Core/z80_cpu.c: In function ‘add_hl_rr’:
Core/z80_cpu.c:341:31: error: suggest parentheses around ‘+’ in operand of ‘&’ [-Werror=parentheses]
     if ( ((unsigned long) hl) + ((unsigned long) rr) & 0x10000) {
          ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
```

Is this the correct logic here?

### DIFF
--- a/Core/z80_cpu.c
+++ b/Core/z80_cpu.c
@@ -338,7 +338,7 @@ static void add_hl_rr(GB_gameboy_t *gb, uint8_t opcode)
         gb->registers[GB_REGISTER_AF] |= GB_HALF_CARRY_FLAG;
     }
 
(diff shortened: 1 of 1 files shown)


LIJI32:
Yes, the logic is correct and more readable this way. Thanks! Similarly to #116, I'd rather this warning disabled (Or alternatively, explicitly enabled for Clang if it's available).

PR #118, opened 2018-10-12
Update libretro dependencies

RobLoach:
This updates the Makefile dependency chain to build rgblink and all when needed. It's missing `flex`, which seems to be a dependency of rgbasm/link, but we'll figure that out later.


RobLoach:
Hmm, not sure about this one. Would love some thoughts....

RobLoach:
@fr500 I've seen you in here a few times. What do you think about building the dependencies through the Makefile like this?

fr500:
I really don't know, I know TA doesn't like dependencies and the current workflow is fine (we have a branch on our fork that I update like... once a month)

RobLoach:
Sounds good.

PR #119, opened 2018-10-13
Don't error out during warnings

RobLoach:
This updates the compilation to not consider all warnings as errors. It will display a warning, but not error out during compilation.


LIJI32:
`-Wall -Werror` is generally a good practice. Any unwanted warnings (e.g. `-Wmultichar`) should be specifically disabled via `-Wno-` (e.g. `-Wno-multichar`)

PR #120, opened 2018-10-13
Ignore multichar and int-in-bool-context warnings

RobLoach:
As mentioned in https://github.com/LIJI32/SameBoy/pull/119#issuecomment-429533870 , this ignores just the two warnings that we run into on GCC. This now allows building in GCC.

@LIJI32 Thanks for the tip on ignoring just those two warnings instead of all :+1:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ endif
 
 # Set compilation and linkage flags based on target, platform and configuration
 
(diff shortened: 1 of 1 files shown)


LIJI32:
Yeah, it's a warning I want to disable so it shouldn't matter if implemented or not. I guess I need to add `-Wno-unknown-warning-option` as well.

LIJI32:
Fixed.

RobLoach:
Thanks for the fix :+1: 

Confirmed working on `gcc` too. Thanks all.

PR #139, opened 2019-01-06
Ready to merge: SGB in the libretro core

fr500:
I tried to hookup sgb but it either crashes or stops at a garbled sgb screen, I think the geometry may be different.

let me know if you have any pointers :)

### DIFF
--- a/libretro/Makefile
+++ b/libretro/Makefile
@@ -6,6 +6,15 @@ ifneq ($(GIT_VERSION)," unknown")
    CFLAGS += -DGIT_VERSION=\"$(GIT_VERSION)\"
 endif
 
+SPACE :=
+SPACE := $(SPACE) $(SPACE)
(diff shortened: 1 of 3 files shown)


fr500:
@LIJI32 this is working fully now and the commits only touch libretro code.

LIJI32:
Does this include support for multiplayer SGB games (like Bomberman)?

fr500:
probably not, I'll check tonight shouldn't be too hard to hook up

fr500:
image

I know my code is a bit redundant, nitpicky but I prefer it like that.
I want to show only the enabled controllers on the frontend, and I want to poll only what is required (should save a few ms here and there that way)

So I set 2p descriptors for linked mode, 1p descriptors for all single modes, and 4p descriptors for SGB

PR #143, opened 2019-01-13
Build portability

bentley:
These three commits allow SameBoy to build on OpenBSD.

* On OpenBSD the compiler doesn’t look in `/usr/local` by default. `pkg-config` is used to provide whatever include directories are necessary for a library.
* SDL’s `pkg-config` files specify an include directory of `${prefix}/include/SDL2`; its headers are meant to be included relative to this directory.
* POSIX specifies the `head(1)` utility, but doesn’t specify the `-c` flag, and OpenBSD doesn’t provide one. `dd(1)` is a portable alternative.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,11 @@ CC := clang
 endif
 endif
 
+# Find libraries with pkg-config if available.
+ifneq (, $(shell which pkg-config))
+PKG_CONFIG := pkg-config
+endif
+
 ifeq ($(PLATFORM),windows32)
 # To force use of the Unix version instead of the Windows version
 MKDIR := $(shell which mkdir)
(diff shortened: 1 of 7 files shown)


orbea:
It would be nice if `pkg-config` wasn't hardcoded and supported alternative implementations like `pkgconf`.

bentley:
You can, with `make PKG_CONFIG=pkgconf`.

orbea:
Yes, you are right, sorry for the confusion.

LIJI32:
I haven't tested this yet, but looking at the changes raises up a few questions/issues:
1. Looks like you're adding the SDL2 and GL CFLAGS even to non-SDL targets (Such as Cocoa and the tester)
2. `SDL2/SDL.h` vs `SDL.h` seems a bit problematic to me. First, I'm not sure this works on macOS without pkgconf. Second, How does this handle SDL1 vs SDL2 assuming both are installed?

LIJI32:
I'm talking about scenarios where pkg-config isn't available, I don't want to require it for targets that already work without it.

bentley:
If pkg-config isn’t available it falls back to the original case.

LIJI32:
But the include statements have changed, and are now ambiguous between SDL1 and 2.

bentley:
Ah, I did overlook that. It’s not that they’re ambiguous, since ${prefix}/include/SDL and ${prefix}/include/SDL2 are not in compiler search paths by default. But by the same token the initial revision won’t build without pkg-config to set search paths; that was unintentional.

I’ve pushed a new revision that falls back to sdl2-config, which comes with SDL2, if pkg-config is not available, and only applies the flags to the SDL target.

bentley:
Any thoughts on this version?

LIJI32:
I kind of forgot about it... 😳

I'll give it a look this weekend.

LIJI32:
It breaks the macOS SDL build when pkg-config is installed – it breaks all SDL includes and doesn't even find libgl (`Package gl was not found in the pkg-config search path.`)

bentley:
That suggests that macOS doesn’t have `gl.pc`—surprising but not unbelievable.

This revision splits out GL build flags from SDL build flags, and sets them from `pkg-config` individually. If `pkg-config` isn’t present or fails to find one library, it falls back to `-lGL` or `sdl2-config`.

Note that even though `GL_CFLAGS` can be blank, it must exist for some systems. On OpenBSD, the libGL headers are under `/usr/X11R6` which is not part of the default compiler search path; `pkg-config` adds that to `GL_CFLAGS`.

I tested this on OpenBSD with `pkg-config` and `gl.pc` present; `pkg-config` present but `gl.pc` absent; `sdl2-config` present but `pkg-config` absent; and got the results I expected.

I moved the bootrom trimming bits to #176 in case this part still needs fine tuning.

twinaphex:
Hi there @LIJI32, any resolution here?

orbea:
@twinaphex I am not sure if you are aware, but this PR should only affect the standalone SameBoy and not libretro.

bentley:
I’ve rebased against master.

If the previous macOS failure is from a missing `gl.pc`, then I believe this fixes it. Like I mentioned previously, I’ve tested with `pkg-config` and `gl.pc` present; `pkg-config` present but `gl.pc` absent; `sdl2-config` present but `pkg-config` absent. Would test on macOS but I don’t have access to any machines that run it.

Any issues with this revision?

LIJI32:
It looks good. I'll test it and merge it.

LIJI32:
It still fails on macOS because it adds `-lGL` in addition to `-framework OpenGL` (The former will not work). I'll fix it later and merge.

bentley:
Thanks!

Issue #111, opened 2018-09-02
compile question: mingw64 gcc 8.1.0 windows libretro


ghost:
Hi, when I compile 32/64-bit, everything works except one spot.
../build/obj/../Core/gb_libretro.c.o:gb.c:(.text+0xcb2): undefined reference to `random'

I did this in gb.c which isn't perchance a smooth idea, but I do get a 88.0 KB dll now and core works great.

#ifndef random
#define random rand
#endif

For future self-compiling, what'd be a better way of avoiding that dangling undefinied reference?

LIJI32:
I'll probably just replace `random` with `rand` entirely, since it causes problems for some other libretro targets as well.

truidic:
Maybe this could work since it only affects libretro?

#ifdef __LIBRETRO__
#define random rand
#endif


Something else I've wondered about is libretro/Makefile. Instead of relying on od and hexdump, would it be more portable/simpler to compile and run your own libretro/hexdump.c that does a plain fopen,fread,printf loop?

cremno:
`srandom()` is never called except in Hexfiend's code. Is not seeding the PRNG intended?

LIJI32:
I didn't really think it mattered if it was deterministic for the first instance, but I guess I should add a `__attribute__((constructor))` function to seed it.

LIJI32:
> Instead of relying on od and hexdump, would it be more portable/simpler to compile and run your own libretro/hexdump.c that does a plain fopen,fread,printf loop?
I'm not sure how libretro's build system works for all different platforms it supports. Since it's more often than not cross-compiling, the build system should be able to build to both the target and the host at once, and I'm not sure the build system actually supports that. `rgbds` is a far heavier dependency anyway.

fr500:
For libretro we build from the libretro fork, buildbot branch with the makefile inside the libretro folder.

ofry:
```
ofry@games MINGW64 ~/main-sameboy
$ make bootroms
cd BootROMs && rgbasm -o ../build/bin/BootROMs/agb_boot.bin.tmp ../BootROMs/agb_boot.asm
rgblink -o build/bin/BootROMs/agb_boot.bin.tmp2 build/bin/BootROMs/agb_boot.bin.tmp
head -c  2304 build/bin/BootROMs/agb_boot.bin.tmp2 > build/bin/BootROMs/agb_boot.bin
[...]

PR #154, opened 2019-02-20
Libretro fix with random() in MSYS2/MinGW

ofry:
Fixes https://github.com/LIJI32/SameBoy/issues/111

Authored by @fr500


fr500:
it's not a fix, it's sweeping the problem under the rug

ofry:
@fr500 Could you suggest another solution?

Issue #157, opened 2019-04-06
Incompatible banked RAM mapping for achievements in libretro (CGB)


rzumer:
Please see [this issue](https://github.com/meleu/RetroArch-problematic-cheevos/issues/39) for reproduction.

It seems like the memory in the banked RAM area (0xD000~0xDFFF) behaves differently from what is expected. Looking at the descriptors being defined in `libretro.c` they look correct at first glance, but as mentioned in the aforementioned issue the contents in memory do not correspond to the expected state, which results in triggers firing when they should not.

Note that it is the current loaded bank that must be pointed to in the map. From what I can tell, SameBoy is pointing to the first bank at all times. I believe that this is the cause of the compatibility issues. Here is the [reference map](https://github.com/RetroAchievements/RAIntegration/blob/master/src/data/ConsoleContext.cpp#L107).

trinemark:
Was just looking at this also earlier but for RA KDL1.
https://github.com/LIJI32/SameBoy/issues/48

Above analysis is correct. D000 (RAM bank x) is not being mapped correctly for either DMG / CGB modes. And 4000 (ROM bank x) somtimes also although I guess it's less used.

I think Sameboy needs more GB_direct_access_t types to handles this correctly.

rzumer:
Oh wow, sorry I did not notice that issue. Glad I could help pin down the issue.

trinemark:
I think I got it. Unlocked Green Scenery for Kirby's Dream Land in both GB/C modes. I'll test Lucky Luke also to be sure.

Reorganized the descs a bit and using 9 of them atm.
```
descs[2].ptr   = GB_get_direct_access(&gameboy[i], GB_DIRECT_ACCESS_RAM, &size, &bank);
descs[2].start = 0xC000;
descs[2].len   = 0x1000;

descs[3].ptr   = descs[2].ptr + (bank * 0x1000);
descs[3].start = 0xD000;
descs[3].len   = 0x1000;

..

descs[6].ptr   = GB_get_direct_access(&gameboy[i], GB_DIRECT_ACCESS_ROM, &size, &bank);
descs[6].start = 0x0000;
descs[6].len   = 0x4000;
descs[6].flags = RETRO_MEMDESC_CONST;

descs[7].ptr   = descs[6].ptr + (bank * 0x4000);
descs[7].start = 0x4000;
descs[7].len   = 0x4000;
descs[7].flags = RETRO_MEMDESC_CONST;
```

Issue #48, opened 2018-03-20
Libretro achievements not working


fr500:
The mmaps are set properly as far as I can see, but achievements are not triggering. The games are identified correctly.

@leiradel I think this worked before right?

fr500:
Ok have some more info

Pokemon Blue achievements didn't unlock when running in GBC mode, they unlocked properly in GB mode
Pokemon Silver achievements didn't unlock in GBC mode, they unlocked properly in GB mode

So it seems to be an issue with GBC

@LIJI32 maybe the offsets are different?

fr500:
Same thing seems to happen in gearboy, he hooked up achievements recently and I guess it's a similar prob

https://github.com/drhelius/Gearboy/blob/master/platforms/libretro/libretro.cpp#L349

He even adds a note about it.

fr500:
Confirmed without a save but by playing this level twice in either mode:

![image](https://user-images.githubusercontent.com/1721040/37640923-a5f197a8-2be5-11e8-8fbf-a7fb34708c5f.png)

LIJI32:
Can you give any example to an achievement definition that does not work on CGB mode, but works on DMG mode?

fr500:
Not sure how the achievements are defined exactly @leiradel would have to help there.
All the achievements I tried last night worked in DMG but not CGB (even for CGB games)

http://retroachievements.org/user/Radius4 see all the GB games on top, I reset all of those and tried in CGB (and never triggered) and DMG (all of those triggered), the pokemon ones triggered because I have SRM.

The one in the screenshot is "beat level 1 in normal difficulty" in Kirby's Dream Land.

leiradel:
I'm sorry, I've been busy with RL...

In this line

```C++
descs[3].ptr = core->IsCGB() ? (core->GetMemory()->GetCGBRAM() + (0x1000 * core->GetMemory()->GetCurrentCGBRAMBank())) : (core->GetMemory()->GetMemoryMap() + 0xD000);
```

What `core->GetMemory()->GetCurrentCGBRAMBank()` returns? This descriptor should always be the RAM bank 1, i.e.

```C++
descs[3].ptr = core->IsCGB() ? (core->GetMemory()->GetCGBRAM() + 0x1000) : (core->GetMemory()->GetMemoryMap() + 0xD000);
```

LIJI32:
That's not SameBoy code. :)

leiradel:
[...]

PR #158, opened 2019-04-06
Fix libretro achievements

trinemark:
https://github.com/LIJI32/SameBoy/issues/48
https://github.com/LIJI32/SameBoy/issues/157


rzumer:
Thank you for the fix.

PR #159, opened 2019-04-07
Add static GBC RAM banks to libretro memory map

rzumer:
This matches a change to the reference map as applied here: https://github.com/RetroAchievements/RAIntegration/commit/6d710b17c33fc1e729eeaf2fbd1bc4c7d011c429.

The region is appended to the tail end to maintain backwards-compatibility.

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -393,7 +393,7 @@ static void init_for_current_model(unsigned id)
             set_link_cable_state(true);
     }
 
-    struct retro_memory_descriptor descs[9];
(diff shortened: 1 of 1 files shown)


rzumer:
Any question/issue with it @LIJI32? It would be great to merge this by next week to match the other libretro cores. The goal is to allow achievements to take advantage of expanded GBC memory, which is necessary for some titles that swap between banks often or in unpredictable ways.

LIJI32:
Sorry, look like I missed this PR. :)

Issue #162, opened 2019-04-12
libretro crash on load when achievements are enabled with some titles


rzumer:
Encountered with "Harry Potter and the Sorcerer's Stone", using the 64-bit Windows version just after @trinemark's PR #158 was merged (so this is not related to #159). I use the GBC bootloader and the crash occurs immediately on load. RetroArch log is not useful at all, there is no error message displayed.

When achievements are disabled, configuration saved and RetroArch restarted, I can launch without an issue, so I think it may be related to the memory descriptors. I don't think that there is enough time to log in and fetch data from the RetroAchievements server before the crash occurs. Other GB and GBC titles that I tested can boot correctly, so it is not a straightforward issue from my perspective. I did not try booting this title in SameBoy before, so I am not sure when the issue started happening, and #158 is not necessarily related.

I also tested other libretro cores that support achievements (Gambatte and Gearboy) and have no problem with those. They have the same memory map as SameBoy currently.

I'm not sure if memory can be accessed before being initialized, see https://github.com/libretro/gambatte-libretro/commit/1a037ec8fc1f68b213925ee1e05ff6cf1bdb5df6 for an example of how to avoid that.

trinemark:
Tried it but couldn't get a crash with Sameboy and Nintendo BIOS. If you can compile the code, try this:
https://github.com/LIJI32/SameBoy/blob/master/libretro/libretro.c#L1055
```
void *retro_get_memory_data(unsigned type)
{
	if (!initialized)
		return 0;

	void *data = NULL;
```

https://github.com/LIJI32/SameBoy/blob/master/libretro/libretro.c#L1120
```
size_t retro_get_memory_size(unsigned type)
{
	if (!initialized)
		return 0;

	size_t size = 0;
```

rzumer:
I don't have an environment set up to build libretro binaries right now, but a silent crash somewhere in the achievements processing pipeline related to the memory info does seem very likely. Looking at the log again, it is clear that the system is loaded partway, but cuts off suddenly with no error output.

rzumer:
Works well, @trinemark. I think this fix can be PRed for safety, but it is likely a symptom of a [RetroArch issue](https://github.com/libretro/RetroArch/issues/8582) where the achievements system fetches the memory info too early and expects it to never change.

rzumer:
A freshly-compiled (debug) build works without an issue, but that's irrelevant of whether the fix is applied or not.

[...]

PR #167, opened 2019-04-26
Update libretro GBC memory map

rzumer:
Matches a change to the reference map (https://github.com/RetroAchievements/RAIntegration/pull/366).

Because RA receives memory layout once and does not change it on the fly, the reference map was not compatible. Recent changes appending GBC banks were revised to only map banks 2-7, while the default bank (1) is set to where the "current bank" used to be (no difference in practice, except for removing a duplicate mapping).

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -414,7 +414,7 @@ static void init_for_current_model(unsigned id)
     descs[2].start = 0xC000;
     descs[2].len   = 0x1000;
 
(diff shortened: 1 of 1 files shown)


rzumer:
Note: this does not attempt to solve #162 (though as mentioned in the issue description, as long as libretro provides an up-to-date build of this core to users with current memory descriptors, it should not occur).

PR #171, opened 2019-05-07
libretro windows compiler (random)

funbars:


funbars:
Notes
https://github.com/LIJI32/SameBoy/issues/111
https://github.com/LIJI32/SameBoy/pull/154

NieDzejkob:
What's the quality of the output of `rand` on Windows? Maybe it'd be a good idea to use something like xorshift (or one of its +/* variants) or PCG, to not rely on the stdlib for rng?

Issue #172, opened 2019-05-08
(Werror) clang 8.0.0 build error


orbea:
OS: `Slackware64-current`
Compiler: `clang-8.0.0`
SameBoy: https://github.com/LIJI32/SameBoy/commit/10be34b5b252711d961ac13a6b8417cb1b94d2bb

I tried to build SameBoy standalone with `clang-8.0.0` and recieved this error.
```
Core/apu.c:69:13: error: loop not unrolled: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Werror,-Wpass-failed=transform-warning]
static void render(GB_gameboy_t *gb, bool no_downsampling, GB_sample_t *dest)
            ^
1 error generated.
make: *** [Makefile:161: build/obj/Core/apu.c.o] Error 1
```
LOG: [SameBoy.log](https://github.com/LIJI32/SameBoy/files/3159053/SameBoy.log)

Maybe it would be good to not force `-Werror`? This just makes it likely to break on new compilers for what might be an otherwise benign warning.

LIJI32:
It's actually a good thing this warning triggered. Unless it fails silently and I'm misremembering something, this optimization does work on my machine and it significantly improves performance. It should be actually fixed and not worked around.

orbea:
Do you know how to fix it? I've never seen this warning before... It would be nice if building SameBoy with clang was also supported. :)

LIJI32:
Well, Clang *is* the recommended compiler to use for SameBoy, and the compiler I use for all official builds (Although I use Apple's fork for macOS, and an older version of it), which is why it's weird. You can disable Werror as a workaround but I'll have to properly investigate it.

orbea:
You're right, I unset the `CC` and `CXX` and SameBoy uses clang by default. Additionally SameBoy unfortunately outright ignores these variables even if I set them to gcc and g++.

Also this warning was introduced sometime after `v0.11.2`.

orbea:
And gcc9 has this issue.
```
Core/display.c:30: error: ignoring #pragma unroll  [-Werror=unknown-pragmas]
   30 |         #pragma unroll
      | 
Core/display.c:46: error: ignoring #pragma unroll  [-Werror=unknown-pragmas]
   46 |         #pragma unroll
      | 
Core/display.c: In function ‘fifo_overlay_object_row’:
[...]

PR #173, opened 2019-05-09
Allow building with gcc.

orbea:
This does two things.

1. It will respect the user set `CC` environment variable if set, it its unset it will default to `clang` like before.
2. It enables pragmas for `gcc` by checking if `__clang__` is defined. This allows building `SameBoy` with gcc without encountering `-Werror` and still allows using them with `clang`.

References:

https://stackoverflow.com/questions/18007326/how-to-change-default-values-of-variables-like-cc-in-makefile
https://stackoverflow.com/questions/28166565/detect-gcc-as-opposed-to-msvc-clang-with-macro

Also see issue https://github.com/LIJI32/SameBoy/issues/172.

### DIFF
--- a/Core/apu.c
+++ b/Core/apu.c
@@ -69,7 +69,9 @@ static void update_sample(GB_gameboy_t *gb, unsigned index, int8_t value, unsign
 static void render(GB_gameboy_t *gb, bool no_downsampling, GB_sample_t *dest)
 {
     GB_sample_t output = {0,0};
+#ifdef __clang__
     #pragma unroll
+#endif
     for (unsigned i = GB_N_CHANNELS; i--;) {
         double multiplier = CH_STEP;
         if (!is_DAC_enabled(gb, i)) {
@@ -126,7 +128,9 @@ static void render(GB_gameboy_t *gb, bool no_downsampling, GB_sample_t *dest)
             unsigned mask = gb->io_registers[GB_IO_NR51];
             unsigned left_volume = 0;
             unsigned right_volume = 0;
+#ifdef __clang__
             #pragma unroll
+#endif
             for (unsigned i = GB_N_CHANNELS; i--;) {
                 if (gb->apu.is_active[i]) {
                     if (mask & 1) {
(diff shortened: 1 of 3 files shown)


LIJI32:
Can you check if `#pragma GCC unroll 8` works as a replacement for `#pragma unroll` in GCC?

orbea:
It builds using `#pragma GCC unroll 8` with gcc-9.1.0. I am not sure how to measure any perfomance measurements.

I noticed now that `make CONF=release` has more `-Werror` errors with gcc9.
```
Core/apu.c: In function ‘update_sample’:
Core/apu.c:62:14: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
   62 |         if (*(uint32_t *)&(gb->apu_output.current_sample[index]) != *(uint32_t *)&output) {
      |              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Core/apu.c:62:70: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
   62 |         if (*(uint32_t *)&(gb->apu_output.current_sample[index]) != *(uint32_t *)&output) {
      |                                                                      ^~~~~~~~~~~~~~~~~~~
Core/apu.c: At top level:
cc1: error: unrecognized command line option ‘-Wno-unknown-warning-option’ [-Werror]
cc1: all warnings being treated as errors
make: *** [Makefile:163: build/obj/Core/apu.c.o] Error 1
```
Its interested that it only recognizes that a command-line option is unrecognized if there are warnings...

LIJI32:
Great, then I'm adding this optimization back to GCC builds (Assuming the gains are similar to Clang's, it's very significant).

Also I could swear I already disabled the strict-aliasing.

LIJI32:
I believe I fixed building with GCC with my last commit, can you verify it please?

orbea:
Yes, both the build with gcc and clang now succeeds. Thanks a lot!

PR #181, opened 2019-05-18
Fix sample lengths for CH1, 2 and 4

ISSOtm:
Unlike what their names suggested, sample lengths attributes don't hold sample lengths, they have to be computed from the attribute instead. Sources:

https://github.com/LIJI32/SameBoy/blob/3ee2c648996c6eb3a2fb32113b984215f54b7b02/Core/apu.c#L428

https://github.com/LIJI32/SameBoy/blob/3ee2c648996c6eb3a2fb32113b984215f54b7b02/Core/apu.c#L445

https://github.com/LIJI32/SameBoy/blob/3ee2c648996c6eb3a2fb32113b984215f54b7b02/Core/apu.c#L465

I feel kinda stupid not to have noticed earlier... :P


LIJI32:
Can you verify these are all in the same units? (APU ticks)

ISSOtm:
Well, they're decremented in the same function (`GB_apu_run`), so they should, unless I'm mistaken.

PR #187, opened 2019-06-27
libretro: Android update

webgeek1234:
Needed for ndk r20

I didn't realize the libretro fork only existed for the buildbot commit until after I had submitted this pr there. Once this is merged here, I'll get someone to clean up downstream.

### DIFF
--- a/libretro/jni/Application.mk
+++ b/libretro/jni/Application.mk
@@ -1,2 +1,2 @@
-APP_STL := gnustl_static 
+APP_STL := c++_static
 APP_ABI := all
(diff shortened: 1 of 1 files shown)


LIJI32:
Should this flag even be there? I don't use *any* C++ code in SameBoy.

webgeek1234:
In that case, no. With no c++ code, the flag should just code away.

LIJI32:
I'm not sure how the Android build system works, will removing this line entirely break anything?

webgeek1234:
Nah, it's not necessary. Several libretro cores are pure ANSI C and don't have that flag.

LIJI32:
Good, so it's gone now. :)

PR #207, opened 2019-10-17
Update from downstream

fr500:
@LIJI32 small updates to the libretro core


LIJI32:
SameBoy doesn't use C++, why would you even need to specify STL?

fr500:
@webgeek1234 can you answer this one please?

webgeek1234:
I don't have time atm to look myself, but I believe if you look back through the PRs here, the android stuff was already talked about and fixed differently from what I had merged to the downstream libretro fork. That's why I said a few times that the fork needed FPed, but I don't think it ever happened.

fr500:
Oh, sorry about that so I'll remove that commit

PR #234, opened 2020-02-27
Spell "length" properly

NieDzejkob:


LIJI32:
Why am I still surprised when I see these? Curse you auto-complete!

PR #241, opened 2020-03-19
SDL: Increase scaling to 4x

lotharsm:
This patch proposes a QoL improvement from my point of view.

Especially on displays with a rather hight screen resolution (HiDPI), the SameBoy window of the SDL version is really small. This patch simply increases the scaling from 2x to 4x.

Another approach could be to use something like a 3x scalier for "standard" DPI screens and 4x for HiDPI screens only, based on the DPI reported by SDL.

### DIFF
--- a/SDL/main.c
+++ b/SDL/main.c
@@ -623,7 +623,7 @@ int main(int argc, char **argv)
     SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
 
     window = SDL_CreateWindow("SameBoy v" xstr(VERSION), SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED,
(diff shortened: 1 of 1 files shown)


lotharsm:
Should I follow this approach more or do you want to dismiss it? :)

LIJI32:
The screen size should *look* on the same DPIs (i.e. 2x on 1x screen scaling and 4x on 2x screen scaling). Isn't it already behaving this way? IIRC that's how it behaved on Windows.

lotharsm:
I can't test it on Windows right now, since the only HiDPI display I have is in my laptop which runs Ubuntu. At least there it has no effect. On Windows, I assume it uses the internal "blurry" scaling provided by Win32 itself? HiDPI-aware apps are quite a mess on Windows...

LIJI32:
SameBoy should handle HiDPI on both Windows 7 and Windows 10 correctly, it shouldn't be blurry.

lotharsm:
Hm... I don't think the `SDL_WINDOW_ALLOW_HIGHDPI` property isn't enough to achieve the scaling, at least from my experience trying to implement HiDPI support for ScummVM. Let me see if I find a test subject.

lotharsm:
(Kinda) superseded by https://github.com/LIJI32/SameBoy/pull/281.

PR #254, opened 2020-05-08
Fix segfault, save some bytes.

NieDzejkob:
Didn't try to compress the palettes yet, but found some bytes to save along the way. Also, the compressor was segfaulting - now it doesn't.


LIJI32:
The segfault is already fixed in master (This branch is no longer updated as it's merged)

NieDzejkob:
Whoops - rebased my PR onto master.

PR #261, opened 2020-05-30
Add ld b, b breakpoint

JL2210:
This is a fairly basic change. Just fire the debugger when `ld b, b` is encountered.

Signed-off-by: James Larrowe <larrowe.semaj11@gmail.com>


LIJI32:
Add a debugger command to enable it, otherwise frontends with debugging disabled might just freeze when they encounter an `ld b,b`

PR #264, opened 2020-06-06
Add SGB Audio support from Bizhawk

nattthebear:
This is an early draft, so just ignore the build script changes and some of the other junkus.  I'm mostly interested in feedback on whether this is something that upstream might be interested in having.

The key details are:
* Blargg's `snes_spc` is used to emulate the audio hardware
* A found-on-the-internet file `sgb-cart-present.spc` is used to provide the initial SPC state
* SGB command `SOU_TRN  = 0x09` is emulated by parsing the resulting packets and `memcpy()`ing them into SPC RAM; the transfer process that would normally involve SNES code is not emulated.
* SGB command `SOUND    = 0x08` is emulated by storing the expected port values and then updating them against the SPC port values at 60hz

It's not much, but it works and games sound fine (once audio output to the frontend is implemented; see below.)

Major missing pieces:
* Useful audio output to the frontend.  Even before this PR, SameBoy's SGB audio output seems to be completely broken, with `GB_sgb_render` blasting chunks of samples directly to the front end while `GB_apu_run` blasts its own chunks of samples directly to the frontend as well, while the frontend then randomly drops chunks of samples because there are too many of them to match the video framerate.  Bizhawk reconciles the timing of the streams together and uses `blip_buf` to resample everything, but I'm not sure how you want to proceed with that.
* Savestates.


LIJI32:
About `sgb-cart-present.spc`, what it contains, how is it created, and what's the legal status of that file? (I.e. does it contain samples from the SGB1/2 ROM that could be protected by copyright?)

nattthebear:
It was most likely made by loading the SNES side SGB rom into a SNES emulator and then dumping the SPC ram and registers after the cart program had filled them all up.  It most certainly contains samples; games that just ask for the built in SGB sound effects to play and don't load their own sample data do work with this.

It's a problem to be sure; it's a firmware but there's not one exact official version of it so it could be quite difficult for users to obtain.  The audio emulation needs the right initial values loaded to the SPC hardware, but without LLE of the SNES side, can't produce that right initial state.

Practically speaking, if you search for spc rips on the internet, it isn't too hard to come across this file, but I can see the issue.

LIJI32:
I see, it's a bit tricky as I'll need to recreate open source version that doesn't use copyrighted samples and code, while keeping it possible for users to provide the original version (Similar to how I handle the boot ROMs).

nattthebear:
Most SNES games used standard sound code, both on the main CPU and the S-CPU side, provided by Nintendo.  The Super Game Boy is no exception, and the behavior I provide in HLE, such as the data packet decoding in `GB_sgb_sound_transfer` and how the HLEed SNES cpu side sets the 4 port values in `GB_sgb_advance_sound`, come from reading a SNES programming manual.

So while there's a lot of reverse engineering ahead of you if you want to incorporate this feature, it's probably not all that "complicated"; the original payload that gets uploaded to the SPC is most likely just standard program data plus a list of patches in some standard format.

PR #268, opened 2020-06-11
Add basic support for 16-bit color output (not tested)

v1993:
Just a quick thing I tried to do. As mentioned in the title, not tested. Please review this if you have time.

As for why, I'm currently trying to port SameBoy onto microcontroller and buffer of `uint32_t` is a serious waste of memory, especially considering that actual color is 16-bit value anyways.

I'm not familiar with internals of SameBoy, so I might have either missed something or changed something I shouldn't had.

### DIFF
--- a/Core/gb_struct_def.h
+++ b/Core/gb_struct_def.h
@@ -1,5 +1,12 @@
 #ifndef gb_struct_def_h
 #define gb_struct_def_h
+#include <stdint.h>
 struct GB_gameboy_s;
 typedef struct GB_gameboy_s GB_gameboy_t;
+
+#ifdef GB_16BIT_OUTPUT_COLOR
+typedef uint16_t GB_output_color_t;
+#else
+typedef uint32_t GB_output_color_t;
+#endif
 #endif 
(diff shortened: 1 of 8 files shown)


Core/gb_struct_def.h
```diff
@@ -1,5 +1,12 @@
 #ifndef gb_struct_def_h
 #define gb_struct_def_h
+#include <stdint.h>
 struct GB_gameboy_s;
 typedef struct GB_gameboy_s GB_gameboy_t;
+
+#ifdef GB_16BIT_OUTPUT_COLOR
```
v1993, line 7: I'm not sure if new `gc_struct_def.h` is a good location for this define, but it seems that it is only working option.

ISSOtm:
Does this belong to upstream? Especially if it's not tested, and no front-end uses it...

LIJI32:
Honestly, I think I replacing `uint32_t` uses with `GB_pixel_t` where relevant is a generally good change, then you can `#ifdef` the definition of that type to allow setting it to some other type. Nothing in the core actually assumes the data size of a pixel as far as I remember. This allows better generalization, and can potentially allow other color depths too, including 8-bit and 24-bit.

PR #270, opened 2020-06-21
Remove NATIVE_CC.

orbea:
This was just wrong so lets revert it...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,10 +16,9 @@ endif
 ifeq ($(PLATFORM),windows32)
 _ := $(shell chcp 65001)
 EXESUFFIX:=.exe
-NATIVE_CC = clang -IWindows -Wno-deprecated-declarations
+CC = clang -IWindows -Wno-deprecated-declarations
 else
 EXESUFFIX:=
-NATIVE_CC := cc
 endif
 
 PB12_COMPRESS := build/pb12$(EXESUFFIX)
@@ -392,7 +391,7 @@ $(OBJ)/BootROMs/SameBoyLogo.pb12: $(OBJ)/BootROMs/SameBoyLogo.2bpp $(PB12_COMPRE
 	$(realpath $(PB12_COMPRESS)) < $< > $@
 	
 $(PB12_COMPRESS): BootROMs/pb12.c
-	$(NATIVE_CC) -Wall -Werror $< -o $@
+	$(CC) $(CFLAGS) -Wall -Werror $(LDFLAGS) $< -o $@
 
 $(BIN)/BootROMs/agb_boot.bin: BootROMs/cgb_boot.asm
 $(BIN)/BootROMs/cgb_boot_fast.bin: BootROMs/cgb_boot.asm


LIJI32:
As said before, it breaks building the libretro core on macOS, and breaking cross-compiling the libretro core. I'm open to other solutions that do not break these workflows. I believe unexporting CC from libretro's Makefile when invoking the parent Makefile in addition to this change *should* fix this issue.

orbea:
And as I said before Libretro should explicitly set $(CC) as needed and compile the build dependency in a separate step using make targets.
```
CC=/path/to/native/compiler make bootroms
cd libretro
make
```
Removing `$(CC)` is a unfortunate hack that is disruptive to every non-broken workflow.

NieDzejkob:
In all the projects I've seen that need to distinguish between the native
and target compiler, the solution was usually a separate environment or
configuration variable, and not a convoluted build procedure.

carmiker:
I will also weigh in here. This build-time complexity has also caused me some grief. I cannot build some boot roms on musl-libc based systems, or at least my Alpine system since the move from pb8 to pb12. I can paste details about that separately, but I think the amount of discussion that has been ongoing on this issue suggests that perhaps taking a few steps back and rethinking this bootrom system could lead to something better. I will not pretend I have the perfect solution, and I respect any decision made on this software whose source code I get for free! Just some food for thought.

orbea:
@NieDzejkob With all due honestly, compiling the build dependency in the same step as the build is a terrible design, but fixing that is a much bigger change. Its better to just get rid of this hack for now. Its insulting that regressions from libretro made it upstream.

NieDzejkob:
Note that the "dependency" is a utility of 100 lines of C code, written for SameBoy, and not a separate software package that has been vendored into the source tree. Could you elaborate on why you consider this a terrible design?

orbea:
The size is irrelevant. Its a dependency because the build process literally depends on it hence the problem with needing it to be built with the host compiler and not the cross compiler.

NieDzejkob:
Okay. Why do you believe that having a single command that updates the build artifacts to reflect the source code is bad design?

orbea:
I'm not sure what you're asking and I'm not sure I care as its far beyond the scope of this PR.

carmiker:
@NieDzejkob, I'm not sure it's possible to minimize the discussion to something so simple without losing the specifics.

Building a build dependency at build time and using realpath to call it later seems a little bit convoluted to me. I'm sure the author has reasons for having designed the project this way, but it is never a bad thing to rethink original design decisions and potentially come up with something closer to ideal when cracks start to show. When earlier I mentioned taking a few steps back, it is because I think discussing how to do the build of a build dependency is not actually the root issue. See: http://xyproblem.info/

What are the advantages of using external bootroms? Is it feasible to make the project buildable in a way that does not require bootroms in cases such as libretro? These are my own questions, which I ask respectfully.

LIJI32:
Libretro still requires the boot ROMs, I do not understand the question. SameBoy provides its own boot ROMS so it can emulate all boot ROM features without providing any Nintendo-authored binaries that may be affected by copyright laws in some countries. The libretro fork manually builds and hardcodes the boot ROM every time I change them, this is not the case in the master branch.

Building project-specific compile time tools in the main makefile is not unheard of, and is very reasonable. Things get complicated when cross-compilation is involved. Now, my requirements for SameBoy's build system are:

1. Single line makefile build. No `./configure`, no `cmake`, no multiple `make` invocations.
2. libretro is a target, and remains a target, including cross compilation. Cross compilation is a common, not a "broken workflow".
3. Cross compiling the SDL build is not supported or an issue right now.
4. Buildable on *at least* macOS, Linux, and Windows with GNUWin
5. No hardcoded tool paths, minimal required dependencies

As I said before, an `unexport` before libretro's make invocation should be enough to fix it, so unless someone has a better idea fulfills my requirements for a build system or wants to implement and test that unexport themselves, I'll do it when I have time for it. Meanwhile, please explain in what way `NATIVE_CC` breaks your workflow so I can avoid it next time.

carmiker:
@LIJI32 thank you, that essentially answers my questions. I will assume that if you are aiming to emulate all boot ROM features via external ROMs, then it is not feasible to make them optional in the future without changing your project goals.

orbea:
> Meanwhile, please explain in what way NATIVE_CC breaks your workflow so I can avoid it next time.

Because users expect `$(CC)` and may not notice when 99% of the build uses the intended compiler and the last 1% uses an unexpected compiler because you are not respecting the correct variable.

> Cross compilation is a common, not a "broken workflow".

Barely supporting osx versions that apple does not support themselves is a broken workflow.

orbea:
Nvm, I changed my mind and uninstalled Sameboy. I don't have the time or energy to argue about stupid things like this.

LIJI32:
As explained more than once, 32-bit macOS build are a specific case of cross compilation, as the libretro core cross compiles to other system such as iOS and Android and various consoles. Your pull request broke those so called "broken workflows", while you haven't explained *why* you would need to explicitly set `$(CC)` and in what way it actually breaks your workflow. You've also been suggested with a fix that would both fix your workflow and keep cross compilation working, but you chose to ignore that as well. If your preferred solution is uninstalling SameBoy, it's fine with me.

PR #271, opened 2020-06-22
Compatibility hacks for old compilers

JL2210:
GCC versions below 4.8.1 didn't have __builtin_bswap16, so provide
a suitable replacement.


LIJI32:
Good catch on the missing -std in the pb12 compilation.

PR #294, opened 2020-10-05
Windows Updates

yo-wotop:
We spoke on Reddit and you helped me get the build process working for Windows.  

I took the various extra command line arguments that we discovered were necessary with the new updates to clang (now defaults to x64 and ignores vcvars32, needs --target) and SDL2 (now needs -lshell32) and added them to the Makefile.  

Also updated the notes within build-faq.md and the readme to clarify dependencies within the Windows build process.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -404,7 +404,7 @@ $(OBJ)/BootROMs/SameBoyLogo.pb12: $(OBJ)/BootROMs/SameBoyLogo.2bpp $(PB12_COMPRE
 	$(realpath $(PB12_COMPRESS)) < $< > $@
 	
 $(PB12_COMPRESS): BootROMs/pb12.c
-	$(NATIVE_CC) -std=c99 -Wall -Werror $< -o $@ --target=i386-pc-windows
+	$(NATIVE_CC) -std=c99 -Wall -Werror $< -o $@
 
 $(BIN)/BootROMs/agb_boot.bin: BootROMs/cgb_boot.asm
 $(BIN)/BootROMs/cgb_boot_fast.bin: BootROMs/cgb_boot.asm
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ endif
 ifeq ($(PLATFORM),windows32)
 _ := $(shell chcp 65001)
 EXESUFFIX:=.exe
-NATIVE_CC = clang -IWindows -Wno-deprecated-declarations
+NATIVE_CC = clang -IWindows -Wno-deprecated-declarations --target=i386-pc-windows
 else
 EXESUFFIX:=
 NATIVE_CC := cc
@@ -129,8 +129,8 @@ GL_CFLAGS := $(shell $(PKG_CONFIG) --cflags gl)
 GL_LDFLAGS := $(shell $(PKG_CONFIG) --libs gl || echo -lGL)
 endif
 ifeq ($(PLATFORM),windows32)
-CFLAGS += -IWindows -Drandom=rand
-LDFLAGS += -lmsvcrt -lcomdlg32 -luser32 -lSDL2main -Wl,/MANIFESTFILE:NUL
+CFLAGS += -IWindows -Drandom=rand --target=i386-pc-windows
+LDFLAGS += -lmsvcrt -lcomdlg32 -luser32 -lshell32 -lSDL2main -Wl,/MANIFESTFILE:NUL --target=i386-pc-windows
 SDL_LDFLAGS := -lSDL2
 GL_LDFLAGS := -lopengl32
 else
@@ -404,7 +404,7 @@ $(OBJ)/BootROMs/SameBoyLogo.pb12: $(OBJ)/BootROMs/SameBoyLogo.2bpp $(PB12_COMPRE
 	$(realpath $(PB12_COMPRESS)) < $< > $@
 	
 $(PB12_COMPRESS): BootROMs/pb12.c
-	$(NATIVE_CC) -std=c99 -Wall -Werror $< -o $@
+	$(NATIVE_CC) -std=c99 -Wall -Werror $< -o $@ --target=i386-pc-windows
 
 $(BIN)/BootROMs/agb_boot.bin: BootROMs/cgb_boot.asm
 $(BIN)/BootROMs/cgb_boot_fast.bin: BootROMs/cgb_boot.asm
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ SameBoy is an open-source project licensed under the MIT license, and you're wel
 SameBoy requires the following tools and libraries to build:
  * clang
  * make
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -404,7 +404,7 @@ $(OBJ)/BootROMs/SameBoyLogo.pb12: $(OBJ)/BootROMs/SameBoyLogo.2bpp $(PB12_COMPRE
 	$(realpath $(PB12_COMPRESS)) < $< > $@
 	
 $(PB12_COMPRESS): BootROMs/pb12.c
-	$(NATIVE_CC) -std=c99 -Wall -Werror $< -o $@
+	$(NATIVE_CC) -std=c99 -Wall -Werror $< -o $@ --target=i386-pc-windows
```
LIJI32, line 407: This affects (and breaks) non-Windows builds, no need for it here since it's already handled in line 19

README.md
```diff
@@ -40,14 +40,14 @@ SameBoy is an open-source project licensed under the MIT license, and you're wel
 SameBoy requires the following tools and libraries to build:
  * clang
  * make
- * Cocoa port: OS X SDK and Xcode command line tools
+ * Cocoa port: OS X SDK and Xcode command line tools [OSX Only]
```
LIJI32, line 43: OS X already implied by the Cocoa, (and the requirement of the OS X SDK and Xcode), no need to put that here

build-faq.md
```diff
@@ -4,4 +4,58 @@ When building on macOS, the build system will make a native Cocoa app by default
...
\ No newline at end of file
+SameBoy on macOS expects you to have SDL2 installed via Brew, and not as a framework. Older versions expected it to be installed as a framework, but this is no longer the case.
+
+# Windows build process
+
+For the various tools and libraries, follow the below guide to ensure easy, proper configuration for the build environment:
+
+#### clang
+
+This may be installed via a Visual Studio installer packages instead of built from source.
```
LIJI32, line 15: Clang can be downloaded from the official LLVM website(https://releases.llvm.org/download.html; "Pre-Built Binaries", Windows 32/64-bit). An official release is better than one provided by a third party.

build-faq.md
```diff
@@ -4,4 +4,58 @@ When building on macOS, the build system will make a native Cocoa app by default
...
+
+#### SDL Port
+
+[libsdl2](https://libsdl.org/download-2.0.php) has two separate files that must be downloaded
+  1. The `-x86` Runtime Binary (e.g., `SDL2-2.0.12-win32-x86.zip` (as of writing))
```
LIJI32, line 20: Why do you need download both? The DLL file is already included in the development zip.

build-faq.md
```diff
@@ -4,4 +4,58 @@ When building on macOS, the build system will make a native Cocoa app by default
...
+- `C:\SDL2\bin\SDL2.dll` will be used as an example
+
+For the Visual C++ Development Library, place the extracted files within a known folder for later.
+
+The following examples will be referenced later:
+
+- `C:\SDL2\lib\x86\*`
+- `C:\SDL2\include\*`
+
+#### Gnuwin
```
LIJI32, line 34: GnuWin

build-faq.md
```diff
@@ -4,4 +4,58 @@ When building on macOS, the build system will make a native Cocoa app by default
...
+
+Ensure that this is in %PATH%.
+
+If errors arise (i.e., particularly with the `CREATE_PROCESS('usr/bin/mkdir')` calls, also verify that Git for Windows has not been installed with full Linux support. If it has, remove `C:\Program Files\Git\usr\bin` from the SYSTEM %PATH% until after compilation.
```
LIJI32, line 38: Can you elaborate about this error? Might be better to actually solve it instead.

build-faq.md
```diff
@@ -4,4 +4,58 @@ When building on macOS, the build system will make a native Cocoa app by default
...
+
+### Building
+
+Within a command prompt in the project directory:
+
+```
+vcvars32
+set path=%path%;C:\SDL2\bin
```
LIJI32, line 46: On my Windows build VM I don't set `%path%`, and SDL2.dll (from the `-devel-` zip) is in the `lib/x86` folder. Can you confirm this works for you? It simplifies the process.

README.md
yo-wotop: I didn't know that Cocoa was OSX-specific when I was first compiling, and when I saw the Windows-specific instructions below, it had implied that all of the above steps were general to all OS's.

I did think that Cocoa might have been OS-specific because I know what Xcode is, but the words "Cocoa port" made me think that there was a "port" of it for Windows. Might be good to break it apart into Linux/Mac/Windows sections to ensure no confusion?

build-faq.md
yo-wotop: Seems reasonable to me. I thought it was between building from source or using VS. I agree with you.

build-faq.md
yo-wotop: I hadn't realized that the DLL was also within the \x86\ folder. Will resolve this and address the %path% concern in the other thread.

build-faq.md
yo-wotop: This works for me as well. I made a mistake in that I thought I needed to add `SDL2.dll` to `%path%`, when I actually needed to add the `rgbds` files to path. I will remove the notes about the SDL2.dll (as well as downloading it separately, to address the other comment) and I will add a section specific to `rgdbgs` path adding.

build-faq.md
yo-wotop: The Git for Windows error, or the `Error -1073741819`?  

For Git for Windows: if you have Git For Windows installed with "Use Git and optional Unix tools from the Command Prompt" ([see SS here](https://i.imgur.com/TMvAojy.png)) selected, a directory containing additional Linux CLI tools will be added to your `SYSTEM PATH` (default: `C:\Program Files\Git\usr\bin`) which contains executables that I believe either trick the installation process into doing Linux-y things on Windows or causes conflicts with default Windows executables (such as `mkdir`). The following error is produced:  

```
process_begin: CreateProcess(NULL, /usr/bin/mkdir -p build/obj/SDL/audio/, ...) failed.
make (e=2): The system cannot find the file specified.
```

However, if you meant the other error:  

> make: *** [build/bin/BootROMs/dmg_boot.bin] Error -1073741819  

This error sometimes occurs during the linking process (typically just before a lot of `cp`s. It is not always reproducible. I have compiled SameBoy about 60-80 times while making edits for some game hacking last week, and I'd say that I ran into this error roughly 70-90% of the time. Simply running `make` a second time solved the issue all-but-1 times, That one final time, I had to run `make` two additional times (3 total).

LIJI32: Aha, it looks like it defaults to the "git" version of `which`, which returns Unix paths instead of Windows paths.

PR #299, opened 2020-10-08
Fix incorrect gamma value.

jkotlinski:


LIJI32:
This `GAMMA` constant doesn't do when you think it does; it's not actually used for gamma correction of the output image. This constant is defined to be specifically the sRGB gamma constant, which is the color space the core uses. The shader uses that constant to convert sRGB to linear RGB, so the scaling shaders can perform color mixing correctly. Then, the same constant is used to convert the scaled image from linear RGB back to the sRGB color space. When using a scaling algorithm that does not introduce new colors to the image, such as Nearest Neighbor or Scale2x, this constant has no effect *at all*.

As for `glEnable(GL_FRAMEBUFFER_SRGB);`, isn't this already the default? Are you sure this had any effect on your machine?

jkotlinski:
Yes, following this change things look good.

LIJI32:
Huh, I see that `GL_FRAMEBUFFER_SRGB` literally does I already do for color correction. Might as well use that and strip that gamma correction from the shader entirely.

jkotlinski:
Yes, and it will look a lot better too.

jkotlinski:
I realize now it is not a problem on all machines. But on those I tested
on, I get results like the screenshot. Post-fix, it looks similar to BGB.

LIJI32:
Can you post a screenshot of how it looks after your fix?

jkotlinski:
image001

LIJI32:
On my machine it completely washes out the colors, even on the menu, and even when color correction is completely disabled.

image

jkotlinski:
Yes, probably the ui colorspace conversion now happens twice, once in CPU
and once in GPU. I would advise to remove the CPU conversion.

LIJI32:
What do you mean by CPU conversion, the color correction code in display.c? It does way more than just Gamma correction, you can't just remove it. Also in the menu itself no CPU conversion actually happens.

LIJI32:
I feel like you're actually fixing the wrong issue. It seems like some of your machines seem to completely skip the CPU color correction. Can you verify that the main screen of SameBoy looks identical to the following screenshot *before* applying your changes?
image

jkotlinski:
It is close, but not identical.
compare

jkotlinski:
I updated the fix to something that works well for me. I find it hard to estimate how well it will work for others.

jkotlinski:
It does not work with all settings - Reduce Contrast is washed out. But in general, this works way better.

Something interesting I found when investigating is that SDL_GetWindowGammaRamp returns an entirely linear curve. Maybe that throws off the SameBoy color correction.

jkotlinski:
I suppose it would take some time and research to figure out the optimal
gamma in this context. BGB has the gamma slider set to 2 by default, but I
suppose those curves might be different.

jkotlinski:
Actually, I can't figure out why it is not working like it should. I will have to give up on this one. Sorry!

PR #320, opened 2021-01-05
add thumbnailer

Jan200101:
This PR adds a basic gnome thumbnailer (similar to QuickLook) to present thumbnails of gameboy roms
image
(image: nautilus showing previews for gameboy roms)

Things that still need figuring out:

- no thumbnailer entry is given
  - this should be solved with a seperate PR that implements a `make install` target

- despite technically working on Windows no explicit support is defined
  - none of the code used is incompatible with Windows but there no work has been done to add support for it in the Makefile

I'd love to get some feedback on this to get the featureset present on MacOS over to Linux

### DIFF
--- a/FreeDesktop/sameboy.thumbnailer
+++ b/FreeDesktop/sameboy.thumbnailer
@@ -1,0 +1,4 @@
+[Thumbnailer Entry]
+TryExec=sameboy-thumbnailer
+Exec=sameboy-thumbnailer "%i" "%o"
+MimeType=application/x-gameboy-rom;application/x-gameboy-color-rom;application/x-gameboy-isx
--- a/Makefile
+++ b/Makefile
@@ -199,6 +199,7 @@ endif
 
 cocoa: $(BIN)/SameBoy.app
 quicklook: $(BIN)/SameBoy.qlgenerator
+thumbnailer: $(BIN)/Thumbnailer/sameboy-thumbnailer
 sdl: $(SDL_TARGET) $(BIN)/SDL/dmg_boot.bin $(BIN)/SDL/cgb_boot.bin $(BIN)/SDL/agb_boot.bin $(BIN)/SDL/sgb_boot.bin $(BIN)/SDL/sgb2_boot.bin $(BIN)/SDL/LICENSE $(BIN)/SDL/registers.sym $(BIN)/SDL/background.bmp $(BIN)/SDL/Shaders
 bootroms: $(BIN)/BootROMs/agb_boot.bin $(BIN)/BootROMs/cgb_boot.bin $(BIN)/BootROMs/dmg_boot.bin $(BIN)/BootROMs/sgb_boot.bin $(BIN)/BootROMs/sgb2_boot.bin
 tester: $(TESTER_TARGET) $(BIN)/tester/dmg_boot.bin $(BIN)/tester/cgb_boot.bin $(BIN)/tester/agb_boot.bin $(BIN)/tester/sgb_boot.bin $(BIN)/tester/sgb2_boot.bin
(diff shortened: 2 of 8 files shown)


LIJI32:
Zlib is perfectly acceptable as a license (Compatible with MIT without requiring a relicense). Personally, I'd like to make SameBoy installable on Linux via compilation, perhaps a Linux `make install` command should be added where by default it installs everything to `/opt/sameboy/`? This should solve the lack of location information while still being overridable so maintainers can customize all of these. This should also allow actual GUI integration (appearing in application menus, having an icon, etc). @NieDzejkob Any opinions?

LIJI32:
Also @max-m previously made his own Gnome thumbnail but it never got actually merged. @max-m Do you remember if your version had any advantages over this iteration? Did it use the icon templates like the Quick Look version?

max-m:
Hi,
my version also used lodepng but also made use of the provided cartridge templates.
I’ve pushed a version that has been rebased onto the latest master to my fork:
https://github.com/max-m/SameBoy/commit/c3844f8750636b27d9c6bdf82030ca553bd45875

I remember that I wasn’t too happy with that code so I never proposed to upstream it.
Feel free to take bits from it as needed :)

Jan200101:
bundling the assets sounds like a good idea and I'll look into using the cartridge templates

Jan200101:
I've ended up reusing the get_image_for_rom_alt implementation though I'd still like feedback on the first point in the original comment.
--boot and --template flags have been added to load a boot rom and template from a specified file.
--size has been added since thumbnailers have a tag for the image size

Jan200101:
> Personally, I'd like to make SameBoy installable on Linux via compilation, perhaps a Linux `make install` command should be added where by default it installs everything to `/opt/sameboy/`?

This is a good idea but perhaps out of the scope of this PR

Jan200101:
hmm, after looking at the code and the code I borrowed from max's fork I'm quite unhappy with the outcome
I'll work on this some more and make it into a draft in the mean time.

Jan200101:
Alright.
- the stb imager resizer has been replaced with a simplistic scaler that doesn't blend colors but gets the job done
- the only cli flags are --boot, to specify what external boot rom to use, and --size, to specify the image size 
- conventions set by CONTRIBUTING.md are followed more closely
- get_image_from_rom.{c,h} are now taken from QuickLook directly to reduce code redundancy

while some of this is not perfect (especially the scaler) I am happy enough with this code

Jan200101:
I am not happy with the scaler anymore and its being reworked

Jan200101:
the general scaling has been simplified now and the output pixels are now being sampled from the input

here is the output of the thumbnailer being ran on Links Awakening with `--size 64` (upscaled for visibility)
image

basxto:
Does it also take an input rom and output image as parameters?

I agree, that the other files are a different issue.
I will open an issue for `make install`.

Jan200101:
Yes it takes input rom and output image as parameters
`./thumbnailer [input rom] [output png]`

Jan200101:
bump

Jan200101:
I can't seem to get rid of the objcopy solution of embedding the files.
As it stands objcopy requires specifying the targe system which is not a good solution so I looked into alternatives and found none short of writing a binary to C command (yes xxd exists but its not readily available on Windows, ld is capable of making binary files into objects but lld isn't without specifying the target, custom solutions like adobe bin2c are not distributed in a lot of distros)

LIJI32:
Sorry, I haven't had time looking at this. Considering this is a Gnome thumbnailer and is not intended to be build on Windows, why is the lack of xxd on Windows even a problem? If and when this code is adapted to work with the relevant Windows APIs, I'll handle that part.

Jan200101:
> Sorry, I haven't had time looking at this.

Yeah no worries

> Considering this is a Gnome thumbnailer and is not intended to be build on Windows, why is the lack of xxd on Windows even a problem?

Now that you mention it there is no good reason for it.
This doesn't aim to be a Windows shell extension so worrying about compiling for it isn't a priority
and those that do want to compile it probably know where to get xxd from

Jan200101:
I completely replaced objcopy with xxd which outputs a C file, cleaned up some of the code to make more sense and added it to the install the thumbnailer to the install target.

Since the install target already puts the boot roms in the correct places I am unsure if embedding them is really needed anymore

Jan200101:
I have decided to move this off into its own project instead
https://github.com/Jan200101/sameboy-thumbnailer

PR #345, opened 2021-02-14
exit with error message instead of crash when a window cannot be opened

jkotlinski:
Quality of life improvement.


LIJI32:
Merged, but where did you actually encounter this problem?

jkotlinski:
I built with Cygwin and forgot to start X.

LIJI32:
Ah.

PR #352, opened 2021-02-27
value_to_string: use snprintf

NieDzejkob:
Currently, value_to_string and debugger_value_to_string use an
error-prone calculation to avoid overflow. This was once adjusted
already, and one of the codepaths is still vulnerable. Put this in a
symfile:

```
01:5678 aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```

and execute `p 1:$5679`. On Linux, the canary terminates the process.

### DIFF
--- a/Core/debugger.c
+++ b/Core/debugger.c
@@ -131,30 +131,25 @@ static const char *value_to_string(GB_gameboy_t *gb, uint16_t value, bool prefer
         symbol = NULL;
     }
 
-    /* Avoid overflow */
-    if (symbol && strlen(symbol->name) >= 240) {
-        symbol = NULL;
-    }
-
     if (!symbol) {
-        sprintf(output, "$%04x", value);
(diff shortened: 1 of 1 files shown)


NieDzejkob:
The size argument to `snprintf` does include the null-terminator, though.

LIJI32:
Right, must have confused with another `n` function, which is why I generally avoid them.

PR #356, opened 2021-02-28
libretro: Sync updates from libretro

RobLoach:
This brings in the latest changes from the libretro core. Includes updates to the build process in the Makefile.

### DIFF
--- a/libretro/Makefile
+++ b/libretro/Makefile
@@ -119,6 +119,20 @@ else ifeq ($(platform), classic_armv7_a7)
 	    LDFLAGS += -static-libgcc -static-libstdc++
 	  endif
 	endif
+
+###########################
+# Raspberry Pi 4 in 64 mode
+else ifneq (,$(findstring rpi4_64,$(platform)))
+   EXT ?= so
+   TARGET := $(TARGET_NAME)_libretro.$(EXT)
+   fpic := -fPIC
+   SHARED := -shared -Wl,--version-script=$(CORE_DIR)/libretro/link.T -Wl,--no-undefined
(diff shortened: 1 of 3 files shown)


libretro/jni/Application.mk
```diff
@@ -1 +1,2 @@
+APP_STL := c++_static
```
LIJI32, line 1: There's no need for this, SameBoy does not use C++.

libretro/Makefile
```diff
@@ -119,6 +119,20 @@ else ifeq ($(platform), classic_armv7_a7)
...
+# Raspberry Pi 4 in 64 mode
+else ifneq (,$(findstring rpi4_64,$(platform)))
+   EXT ?= so
+   TARGET := $(TARGET_NAME)_libretro.$(EXT)
+   fpic := -fPIC
+   SHARED := -shared -Wl,--version-script=$(CORE_DIR)/libretro/link.T -Wl,--no-undefined
+   CFLAGS += -O2 -march=armv8-a+crc+simd -mtune=cortex-a72
+   CXXFLAGS += $(CFLAGS)
+   CPPFLAGS += $(CFLAGS)
+   ASFLAGS += $(CFLAGS)
```
LIJI32, line 133: No need for AS/CXX/CPP flags, they're not used.

libretro/jni/Application.mk
RobLoach:
``` suggestion
```

libretro/Makefile
RobLoach: Thanks!

PR #370, opened 2021-03-25
replace PREFIX/share/sameboy with DATA_DIR

Jan200101:
installation does not respect DATA_DIR which can cause a mismatch between what the code expects 
https://github.com/LIJI32/SameBoy/blob/925bd863c08334b54cd075b6da9e537688ecb76b/Makefile#L57
and where the files actually are


LIJI32:
Good catch!

PR #393, opened 2021-06-26
[Libretro] Upgrade Core Options to v1.3 with sublabels and translation support + expose settings

Ryunam:
This PR upgrades the core options in the libretro implementation of SameBoy to version 1.3, allowing for descriptions of each option in the sublabels, more detailed descriptions of each value as well as translation support. See here for further reference: https://github.com/libretro/libretro-common/tree/master/samples/core_options

All the changes implemented with this PR are libretro-specific and stay contained within the corresponding files. I also took the chance to expose a few settings that have existed in SameBoy for quite some time, but had yet to be made configurable in the libretro core.

Here is a list of the main changes:

- I have added the `retro_inline.h` file from the `libretro-common` repository, which is a requirement for the v1.3 core options.
- `libretro_core_options_intl.h` was added as well. This is the file that will contain any future translation work for the core options text.
- The `libretro.h` header file was replaced with the most up-to-date version.
- As per the specification, `libretro_core_options.h` now contains the text for all the labels, sublabels and values of the core options available for SameBoy.
- A few instances of unneeded whitespace were cleaned up in `libretro.c`
- Several edits were applied to `libretro.c` to adhere to the v1.3 format. For example, the original code that was used to separate the "single cartridge mode" variables from the "dual cartridge mode" variables was not adequate for the new implementation, so it was replaced with a function that checks the internal name of each variable and assigns them to either list, then makes one or the other list visible based on the number of emulated devices that are being run at any given time. Credits go to the Mupen libretro core for this idea: https://github.com/libretro/mupen64plus-libretro-nx/blob/b4024d75597da162bbc7cb693b8cef0d6da75105/libretro/libretro.c#L258
- The options were reordered by strictly following this sequence: System -> Video -> Audio -> Input.

---

The following options that were not visible in the libretro core are now exposed:

1. The "Game Boy Color (CPU-CGB C) (Experimental)" option for the emulated model is now selectable.

2. The "Mono Palette" for content played on the original Game Boy model is now configurable.

3. The "Ambient Light Temperature" option is now available with 21 values. Default is "6500K (Neutral White)".

4. The "Interference Volume" option is also now available in the libretro core, albeit with an important caveat. This is the only aspect where I am unsure of the outcome: at 100% the volume of the buzzing effect in the libretro core sounds noticeably lower compared to the same interference volume in the standalone Windows version. The option in itself seems to work, but it might be that I made a mistake with the values, so I would appreciate some assistance in order to figure this out more accurately.

5. The same options mentioned here are also available in "dual cartridge mode", with a distinction between the option that affects Game Boy n.1 and the other identical option that affects Game Boy n.2.

---

This was tested on Windows 10 x64. Here's a sample picture of how the core options look in single cartridge mode:

sameboy-coreoptions-1 3

This is how they look in dual cartridge mode:

sameboy-coreoptionsdual-1 3

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -22,6 +22,7 @@
 
 #include <Core/gb.h>
 #include "libretro.h"
+#include "libretro_core_options.h"
 
 #ifdef _WIN32
 static const char slash = '\\';
@@ -45,20 +46,24 @@ char battery_save_path[512];
 char symbols_path[512];
 
 enum model {
-    MODEL_DMG,
-    MODEL_CGB,
+    MODEL_DMG_B,
+    MODEL_CGB_C,
+    MODEL_CGB_E,
     MODEL_AGB,
-    MODEL_SGB,
+    MODEL_SGB_PAL,
+    MODEL_SGB_NTSC,
     MODEL_SGB2,
     MODEL_AUTO
 };
 
 static const GB_model_t libretro_to_internal_model[] =
 {
-    [MODEL_DMG] = GB_MODEL_DMG_B,
-    [MODEL_CGB] = GB_MODEL_CGB_E,
+    [MODEL_DMG_B] = GB_MODEL_DMG_B,
+    [MODEL_CGB_C] = GB_MODEL_CGB_C,
+    [MODEL_CGB_E] = GB_MODEL_CGB_E,
     [MODEL_AGB] = GB_MODEL_AGB,
-    [MODEL_SGB] = GB_MODEL_SGB,
+    [MODEL_SGB_PAL] = GB_MODEL_SGB_PAL,
+    [MODEL_SGB_NTSC] = GB_MODEL_SGB_NTSC,
     [MODEL_SGB2] = GB_MODEL_SGB2
 };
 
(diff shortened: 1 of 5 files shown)


RokkumanX:
Way to go brother, hell yeah that looks nice!

LIJI32:
I removed `libretro_core_options_intl.h` for now, as I don't see translation support being added any time soon. I merged it after a few minor fixes and confirmed it's working on macOS.

Ryunam:
Thank you! I was actually thinking of working on a full Italian translation of the core options upon approval of this PR, so let me know if that would be considered adequate as a first contribution to translation support.

LIJI32:
I see. I've always been a bit of hesitant about translation support in the various flavors of SameBoy due my inability to verify the majority of translations or being able to update them if I eventually update/add/clarify the English strings. I guess it's not that bad for the libretro version as the only strings are the options.

Ryunam:
Yes, that is of course an understandable concern. For what it's worth I am a native Italian speaker, so for this particular translation effort that I was planning to attempt I believe I can do a very accurate localization of the strings. Anyway, I'll try to submit a PR soon so that you can take a general look at the end result.

PR #429, opened 2022-01-14
Silence more -pedantic warnings

orbea:
This fixes two more warnings with `-pedantic` in gb.h.

1.
```
gb.h:400:2: warning: embedding a directive within macro arguments has undefined behavior [-Wembedded-directive]
#ifdef GB_BIG_ENDIAN
 ^
gb.h:410:2: warning: embedding a directive within macro arguments has undefined behavior [-Wembedded-directive]
#endif
 ^
```
2.
```
gb.h:731:2: warning: embedding a directive within macro arguments has undefined behavior [-Wembedded-directive]
#define GB_REWIND_FRAMES_PER_KEY 255
 ^
```

### DIFF
--- a/Core/gb.h
+++ b/Core/gb.h
@@ -44,6 +44,11 @@
 #error Unable to detect endianess
 #endif
 
+#ifdef GB_BIG_ENDIAN
+#define GB_REGISTER_ORDER a,f,b,c,d,e,h,l;
+#else
+#define GB_REGISTER_ORDER f,a,c,b,e,d,l,h;
+#endif
 
 typedef struct {
     struct GB_color_s {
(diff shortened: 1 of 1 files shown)


Core/gb.h
```diff
@@ -44,6 +46,11 @@
 #error Unable to detect endianess
 #endif
 
+#ifdef GB_BIG_ENDIAN
+#define GB_REGISTER_ORDER(a,b,c,d,e,f,h,l) a,f,b,c,d,e,h,l
```
LIJI32, line 50: This macro would be clearer without arguments (It's rather confusing this way)

orbea: I changed it, I think this is better?

PR #430, opened 2022-01-14
sgb.h: Silence -Wgnu-anonymous-struct and -Wnested-anon-type warnings

orbea:
When including `gb.h` as a dependency in a C++ project even with `extern "C"` warnings like this are exposed with `-pedantic`. I'm not really sure what to call this union and struct so please suggest better names if you want.
```
sgb.h:14:9: warning: anonymous structs are a GNU extension [-Wgnu-anonymous-struct]
        struct {
        ^
sgb.h:14:9: warning: anonymous types declared in an anonymous union are an extension [-Wnested-anon-types]
        struct {
        ^
```

### DIFF
--- a/Core/sgb.h
+++ b/Core/sgb.h
@@ -7,6 +7,7 @@
 typedef struct GB_sgb_s GB_sgb_t;
 typedef struct {
     uint8_t tiles[0x100 * 8 * 4];
+#ifdef GB_INTERNAL
     union {
         struct {
             uint16_t map[32 * 32];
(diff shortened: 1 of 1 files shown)


LIJI32:
Adding `sgb_u.sgb_s.` to every `GB_sgb_border_t` access isn't really acceptable. It's better to just replace the anonymous union with just `uint16_t raw_data[0x440];` under a `__cplusplus` ifdef, as internal structs data shouldn't be accessed directly outside of the core, which is C anyway.

orbea:
> as internal structs data shouldn't be accessed directly outside of the core

Probably best to add `GB_INTERNAL` then? As that would side step the issue for me I did that, but I can add `__cplusplus` also if you think its still desirable?

LIJI32:
`GB_INTERNAL` works just as fine.

PR #431, opened 2022-01-20
[SDL] Fix loading of the configured DMG palette

max-m:
There are four options for the DMG color palette:
- Greyscale
- Lime (Game Boy)
- Olive (Pocket)
- Teal (Light)

When loading the configuration from the `prefs.bin`
the configured `dmg_palette` gets normalized by applying
`mod 3` to it, limiting the possible values to [ 0, 1, 2 ],
thus selecting the “Teal” color palette will be saved as such,
but the next time the config gets loaded it gets reset to Greyscale.


LIJI32:
This is what happens when I think & and write % 🦆

PR #433, opened 2022-02-02
Don't flash (open/close) a console on Windows

sredna:
The console code tries to determine if it has a real console by calling `AllocConsole` but this will briefly show a console window in the non-debugger version. `GetConsoleMode` does not work on redirected handles, it is a much better choice.

This only checks if stdout is connected to a real console, it does not check stdin on purpose. This should in theory allow something like `sameboy_debugger.exe < commands_to_execute.txt`.

### DIFF
--- a/SDL/console.c
+++ b/SDL/console.c
@@ -85,28 +85,30 @@ static bool is_term(void)
 {
     if (!isatty(STDIN_FILENO) || !isatty(STDOUT_FILENO)) return false;
 #ifdef _WIN32
-    if (AllocConsole()) {
-        FreeConsole();
-        return false;
+    unsigned long input_mode, output_mode, has_con_input, has_con_output;
+    
+    HANDLE hSI = GetStdHandle(STD_INPUT_HANDLE);
+    HANDLE hSO = GetStdHandle(STD_OUTPUT_HANDLE);
+    
+    has_con_input = GetConsoleMode(hSI, &input_mode);
+    has_con_output = GetConsoleMode(hSO, &output_mode);
+    if (!has_con_output) {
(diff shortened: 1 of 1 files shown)


sredna:
Just in case there is some doubt about using this technique, MSDN [says](https://docs.microsoft.com/en-us/windows/console/writeconsole): 

> ...determine whether the output handle is a console handle (one method is to call the `GetConsoleMode `function and check whether it succeeds).

SDL/console.c
```diff
@@ -85,28 +85,30 @@ static bool is_term(void)
...
     
-    unsigned long input_mode, output_mode;
+    HANDLE hSI = GetStdHandle(STD_INPUT_HANDLE);
+    HANDLE hSO = GetStdHandle(STD_OUTPUT_HANDLE);
     
-    GetConsoleMode(GetStdHandle(STD_INPUT_HANDLE), &input_mode);
-    GetConsoleMode(GetStdHandle(STD_OUTPUT_HANDLE), &output_mode);
-    SetConsoleMode(GetStdHandle(STD_INPUT_HANDLE), ENABLE_VIRTUAL_TERMINAL_INPUT);
-    SetConsoleMode(GetStdHandle(STD_OUTPUT_HANDLE), ENABLE_WRAP_AT_EOL_OUTPUT | ENABLE_PROCESSED_OUTPUT | ENABLE_VIRTUAL_TERMINAL_PROCESSING);
+    has_con_input = GetConsoleMode(hSI, &input_mode);
```
ISSOtm, line 93: AFAICT this is not used?

ISSOtm (changes requested):

PR #436, opened 2022-02-15
BMP Screenshots for linux

mechanicallamb:
Implementation of screenshots for linux users only (someone else will have to implement the directory creation on windows, as I cant get SDL [and by extension sameboy] to build). MAC is on its own. Attempting to take screenshots with non-linux machines will not modify the system

Ctrl+S while playing creates a screenshot directory in the same directory as the rom and .sav. A new directory is created for each rom. Screenshots are saved as [ROM_TITLE]+[base 10(32-bit number)].bmp

This was tested on Ubuntu 20.04 with games pokemon crystal, legend of zelda: oracle of seasons, and legend of zelda: a link to the past


ISSOtm:
This is a good idea, but after talking with @LIJI32, there is an agreement that using BMP is Not Great&mdash;PNG would be preferable instead. Consider using libpng instead?

(Using libpng may end up being necessary for other purposes, such as a Linux thumbnailer. I know Liji doesn't like its API, but I'm volunteering to co-own that kind of code.)

Also, it is highly ill-advised to using fixed-size buffers for a lot of the operations you're doing, especially path manipulation. This creates a lot of vulns from long path names, and while SameBoy isn't security-critical software, it's nice to avoid crashes where possible, and also potentially being a backdoor vector (unlikely as that may be...).

Please feel free to ask for advice on how to implement any of the above, fwiw.

PR #444, opened 2022-04-01
Libretro: Batch upload audio to frontend after video upload

realnc:
Uploading audio at the end of retro_run() with retro_audio_sample_batch_t instead of using retro_audio_sample_t to submit samples as they are generated has peformance benefits when using low audio latency settings. Frontends that implement audio sync and DRC (like RetroArch) will block for a long time when submitting audio. This can cause video frame presentation to be delayed too long. Low audio latency settings make it worse, as DRC and audio sync is performed more often, creating a bottleneck.

Simply accumulating all generated audio samples and uploading them at the end with retro_audio_sample_batch_t improves things condiderably.

On a Linux system with PipeWire as the sound system, and RetroArch configured with an audio latency of just 5ms using the ALSA driver (the lowest latency my audio hardware can do reliably,) this is how the core sounds prior to this commit:

https://user-images.githubusercontent.com/1665903/161177371-1727e4a2-9137-4a79-81e7-92beedd91504.mp4

And this is how it sounds with this commit:

https://user-images.githubusercontent.com/1665903/161177593-e081f290-11e3-49b3-9b87-21971aaff8e0.mp4

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -86,7 +86,7 @@ static struct retro_log_callback logging;
 static retro_log_printf_t log_cb;
 
 static retro_video_refresh_t video_cb;
-static retro_audio_sample_t audio_sample_cb;
+static retro_audio_sample_batch_t audio_batch_cb;
 static retro_input_poll_t input_poll_cb;
 static retro_input_state_t input_state_cb;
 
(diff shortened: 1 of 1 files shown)


realnc:
I've marked it as a draft for now, since it turns out that RetroArch only recently gained the ability to handle that many samples at once (384 kHz is a lot of samples.) So this change might break other frontends.

realnc:
Marked as ready again. The audio upload now checks how many audio frames were actually sampled by the frontend. It will keep uploading until all audio has been sampled.

PR #450, opened 2022-04-18
use friendly names for HW registers instead of magic numbers

remind-me-later:
Use the definitions from [hardware.inc](https://github.com/gbdev/hardware.inc) to assign names to the hardware registers in the boot ROMs.


remind-me-later:
Glad I could help :smile:

PR #453, opened 2022-05-21
fix rlc_r zero flag

remind-me-later:
The comparison to set the zero flag in rlc_r discards the carry bit, it somehow worked before so maybe there's something I missed. Anyways this is slightly simpler.


LIJI32:
It worked because the left shift casts it to an `int`. In any case your solution is definitely cleaner.

PR #454, opened 2022-05-21
fix inc_lr and dec_lr overflow

remind-me-later:
I think the old version could overflow changing the value of the high register, but worked fine. I'm sorry if this is another instance of C doing some implicit casting I'm not seeing.

### DIFF
--- a/Core/sm83_cpu.c
+++ b/Core/sm83_cpu.c
@@ -595,7 +595,7 @@ static void inc_lr(GB_gameboy_t *gb, uint8_t opcode)
     uint8_t value;
     register_id = (opcode >> 4) + 1;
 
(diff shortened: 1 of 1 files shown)


Core/sm83_cpu.c
```diff
@@ -595,7 +595,7 @@ static void inc_lr(GB_gameboy_t *gb, uint8_t opcode)
     uint8_t value;
     register_id = (opcode >> 4) + 1;
 
-    value = (gb->registers[register_id] & 0xFF) + 1;
+    value = (gb->registers[register_id] + 1) & 0xFF;
```
jkotlinski, line 598: My guess is that the & 0xff part is redundant.

Issue #226, opened 2020-01-12
Poor audio on OpenBSD


bentley:
When playing games on OpenBSD with the SDL2 frontend, audio has lots of clicks and noises, and emulation is noticeably slower than it should be.

Attached are audio recordings of (version 0.12.3 of) the SDL frontend and of the libretro core running under RetroArch, which does not have these issues. The clicks in the SDL recording are representative of what you hear during normal gameplay. The clicks in the libretro recording seem to be artifacts of the recording—nothing similar is audible when playing normally.

[openbsd-audio.zip](https://github.com/LIJI32/SameBoy/files/4050526/openbsd-audio.zip)

LIJI32:
Sadly, if it only reproduces on OpenBSD and not other frontends, it's most likely an SDL2 bug that I pretty much don't have any means to fix. SDL2 audio turned out to be *very* buggy over SameBoy's development, with newer versions of SDL2 breaking some audio configurations on macOS and Linux, but I don't know any better cross-platform alternative. Can you try with different (Newer or older) versions of SDL2 and see if it has any effect? If so, it might be more effective to report it the SDL folks.

LIJI32:
I pushed a branch with a port-audio driver for the SDL frontend. Sadly, it seems to be completely broken on ALSA on Linux, but maybe it performs better on OpenBSD. Let me know.

bentley:
Thanks for checking. I built and ran the portaudio and SDL audio backends, but couldn’t distinguish any difference in sound quality between them: both had serious clicking along with slow emulation.

There were two differences in the portaudio version: bringing up the menu with Esc had noticeable delay, and hitting ^C during emulation triggered a crash:

```
#0  0x00000d6071d63510 in memcpy (dst0=<optimized out>, src0=<optimized out>, 
    length=3840) at /usr/src/lib/libc/string/memcpy.c:103
#1  0x00000d5e3cd2f321 in callback (in=0x0, _out=0xd611131b100, frames=960, 
    time_info=0xd60b0eab108, status_flags=0, unused=0x0)
    at SDL/audio/portaudio.c:113
#2  0x00000d6107676f41 in NonAdaptingProcess ()
   from /usr/local/lib/libportaudio.so.1.2
#3  0x00000d6107676927 in PaUtil_EndBufferProcessing ()
   from /usr/local/lib/libportaudio.so.1.2
#4  0x00000d610767a0d3 in sndioThread ()
[...]

PR #471, opened 2022-07-09
Add an optional OpenAL audio driver for the SDL frontend

max-m:
~~To compile the OpenAL driver specify `ENABLE_OPENAL=1` when invoking `make`.~~
I chose to make it optional because I didn’t want to force everyone to install OpenAL development headers.
Edit: The latest commit changes this behavior slightly. Now the driver gets compiled if OpenAL is detected via `pkg-config`. If `pig-config` is unavailable you can enable the usage of OpenAL like before and use `ENABLE_OPENAL=0` to disable it in all cases. 

Because my system complained with `AL lib: (EE) alc_cleanup: 1 device not closed` when I quit SameBoy I added the `audio_deinit` function to the driver interface and registered it via `atexit()`.

I’ve tested the backend on my desktop running Ubuntu 20.04 with PipeWire as audio server as well as my WASM fork of SameBoy.
The native build on my desktop didn’t crackle or exhibit any other typical audio-related issues, so I see this as a win. :)
~~At the moment the `BUFFER_LEN_MS` is set to a rather conservative 32 milliseconds (if my math is right), but I think this could be lowered when targeting modern systems. Edit: PortAudio for example reports an output latency of \~11ms on my system.~~

I hope that I didn’t make any stupid OpenAL-mistakes because I was completely unfamiliar with the [API](https://www.openal.org/documentation/OpenAL_Programmers_Guide.pdf).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,12 +24,6 @@ NATIVE_CC := cc
 SDL_AUDIO_DRIVERS ?= sdl
 endif
 
-ifneq ($(ENABLE_OPENAL),)
-CFLAGS += -DENABLE_OPENAL
-LDFLAGS += -lopenal
-SDL_AUDIO_DRIVERS += openal
-endif
-
 PB12_COMPRESS := build/pb12$(EXESUFFIX)
 
 ifeq ($(PLATFORM),Darwin)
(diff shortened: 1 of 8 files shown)


max-m:
This might help with #226, in case this issue is still relevant.

max-m:
I’ve pushed a couple of small changes to the initial commit. If you want I can squash them, I just split them into multiple commits for clarity.

max-m:
I’ve rebased the PR, incorporated the changes to `_audio_get_queue_length()` from a856566 and reduced the internal buffer size down to 5 ms, like the SDL frontend.

Issue #491, opened 2022-08-27
Flathub


Mailaender:
Hi, I created a Flatpak package at https://github.com/flathub/flathub/pull/3442.

Mailaender:
https://flathub.org/apps/details/io.github.sameboy.SameBoy

PR #492, opened 2022-08-29
Added AppStream metadata

Mailaender:
required for #491. https://www.freedesktop.org/software/appstream/docs/ explains what this is about. I didn't count for https://hughsie.github.io/oars/ as the ROMs contain the data and are not shipped with the emulator.


LIJI32:
Few questions/notes:
1. What is the purpose of the new Makefile rule? No target seems to require it so it won't ever be executed and copied over.
2. Isn't this metadata for repositories themselves? Why would it need to be installed anyway?
3. `<metadata_license>CC0-1.0</metadata_license>` is problematic. I didn't license anything in my repo under a CC license
4. It's better if the screenshots are hosted in-repo rather than some 3rd-party website. They're outdated anyway (0.12.3, old menu background, old menu contents, old boot logo, old color correction)

Mailaender:
1. I was confused by the Makefile structure as I have never seen someone do file based targets. Where do I add it?
2. https://freedesktop.org/software/appstream/docs/chap-Metadata.html is meant to be for upstream inclusion.
3. metadata license is the license for the XML file not your code
4. I am also not quite happy about the Ubuntu artwork in the screenshot as it should be neutral. The ones on your homepage seem to be macOS exclusive, though.

QuLogic:
> I was confused by the Makefile structure as I have never seen someone do file based targets. Where do I add it?

The `install` target should depend on it, like it does for the icons.

PR #502, opened 2022-10-10
Fix -Werror=strict-prototypes

orbea:
This will be required for upcoming gcc and clang versions.

Reference: https://archives.gentoo.org/gentoo-dev/message/dd9f2d3082b8b6f8dfbccb0639e6e240

### DIFF
--- a/BootROMs/pb12.c
+++ b/BootROMs/pb12.c
@@ -25,7 +25,7 @@ void write_all(int fd, const void *buf, size_t count) {
     }
 }
 
-int main()
+int main(void)
 {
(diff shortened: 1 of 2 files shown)


LIJI32:
Technically I that `nop()` was intended as I wanted to use the "unspecified arguments" C behavior, but it didn't really matter.

orbea:
Thanks for the quick review!

> Technically that `nop()` was intended as I wanted to use the "unspecified arguments" C behavior, but it didn't really matter.

gcc didn't care about that one, but clang was being more strict.

PR #521, opened 2022-12-30
[CI] Update to actions/checkout@v3

max-m:
See https://github.com/actions/checkout/issues/1047


max-m:
I’ve force pushed over my previous commit to include the newly added `.github/workflows/libretro.yml` as well.

LIJI32:
Not sure how exactly git broke, but this PR is merged (Despite not appearing so)

PR #525, opened 2023-02-07
Fix MGB bootrom padding larger than 256 bytes

deltabeard:
mgb_boot.bin was incorrectly being padded to a CGB-size bootrom.

Signed-off-by: Mahyar Koshkouei <mk@deltabeard.com>


LIJI32:
Good catch!

Issue #399, opened 2021-08-06
read ZIP format?


clavinet:
It would be neat if SameBoy could read zips by itself, unless that is out of scope in which case feel free to close this.

PR #535, opened 2023-03-29
Enable loading zipped rom files

brianmwaters:
Solves https://github.com/LIJI32/SameBoy/issues/399, and I think it's a useful feature I plan to use locally.

This adds a new dependency ([minizip](https://www.winimage.com/zLibDll/minizip.html)), some tests in the makefile, and a little bit of ifdef spaghetti. I added an `ENABLE_MINIZIP` variable to make this feature optional, but I defaulted it to 1.

I tried to be careful in editing the makefile, but I haven't tested on Windows or OS X. Minizip is very small (just a single .c/.h pair), so the best solution might be to just copy-paste the entire thing into the project and enable the feature unconditionally, which would get rid of the makefile checks and the ifdef spaghetti.

It does feel like this change belongs outside of `Core/gb.c`, for example in `SDL/main.c` where `GB_load_rom` is called. However, code would have to be changed at every call site of `GB_load_rom`, and there are a handful of those. Instead, this just makes `GB_load_rom` "do the right thing" transparently, and adds a `GB_load_rom_from_bin` in case anybody ever needs to access the old behavior. I am open to refactoring this patch to work some other way, though.

### DIFF
--- a/Core/gb.c
+++ b/Core/gb.c
@@ -10,7 +10,7 @@
 #include <unistd.h>
 #endif
 
-#ifdef HAVE_MINIZIP
+#ifdef GB_HAS_MINIZIP
 #include <minizip/unzip.h>
 #endif
 
@@ -335,7 +335,7 @@ static void init_new_rom(GB_gameboy_t *gb)
 
 int GB_load_rom(GB_gameboy_t *gb, const char *path)
 {
-#ifdef HAVE_MINIZIP
+#ifdef GB_HAS_MINIZIP
     size_t path_len = strlen(path);
     const char *ext;
     if (path_len > 4) {
@@ -373,7 +373,7 @@ int GB_load_rom_from_bin(GB_gameboy_t *gb, const char *path)
     return 0;
 }
 
-#ifdef HAVE_MINIZIP
+#ifdef GB_HAS_MINIZIP
 int GB_load_rom_from_zip(GB_gameboy_t *gb, const char *path)
 {
     GB_ASSERT_NOT_RUNNING_OTHER_THREAD(gb)
--- a/Core/gb.c
+++ b/Core/gb.c
@@ -353,19 +353,21 @@ int GB_load_rom_from_bin(GB_gameboy_t *gb, const char *path)
     GB_ASSERT_NOT_RUNNING_OTHER_THREAD(gb)
     
     FILE *f = fopen(path, "rb");
+    size_t bin_size;
     if (!f) {
         GB_log(gb, "Could not open ROM: %s.\n", strerror(errno));
         return errno;
     }
     fseek(f, 0, SEEK_END);
-    gb->rom_size = rounded_rom_size(ftell(f));
+    bin_size = ftell(f);
+    gb->rom_size = rounded_rom_size(bin_size);
     fseek(f, 0, SEEK_SET);
     if (gb->rom) {
         free(gb->rom);
     }
     gb->rom = malloc(gb->rom_size);
     memset(gb->rom, 0xFF, gb->rom_size); /* Pad with 0xFFs */
-    fread(gb->rom, 1, gb->rom_size, f);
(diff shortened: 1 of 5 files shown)


brianmwaters:
The second commit fixes a subtle issue in the master branch, where SameBoy requests to `fread` the rounded-up number of bytes from the rom file. This silently succeeds because the return value of `fread` is not checked.

I blindly copied the same pattern with my zip patch, but later realized the issue. I am not sure what the behavior of minzip is in this case (it might overwrite the extra space w/ junk data?)... I didn't bother to even check, and opted to just make the code do the "right thing" instead.

LIJI32:
Few notes:
* I think it would be better to have this code as part of the SDL frontend rather than the core itself. The current changeset also breaks the non-SDL executable targets (Cocoa, iOS, tester...) because they're not linked against minizip
* Use `GB_HAS_MINIZIP` rather than `HAVE_MINIZIP` to not pollute the global "namespace"
* Might want to automatically set it to 0 if minizip is not found
* Last one's a bit tricky; on Windows, minizip's DLL is expected to be copied to bin/SDL folder (similarly to SDL2.dll). Additionally, the Windows release will redistribute that DLL, minizip's license needs to be appended to the bin/SDL/LICENSE file.

brianmwaters:
Sounds good, I will work on that.

I don't understand your last bullet point, though. (I haven't hset up a GnuWin32 build environment yet, so I'm not familiar with how everything is laid out there.) Are you green-lighting my suggestion to bring the latest minizip into the tree, rather than link against it as an external dependency?

LIJI32:
Having a copy of minizip in the tree would sure simplify things (just make sure to update `.gitattributes` and mark it as `linguist-vendored`), so only the license issue remains for that bullet.

brianmwaters:
Made some minor changes as you suggested, but I did not refactor the zipped rom loading code into SDL/.

While I was working on this I realized a few things:

- As it stands, Cocoa and iOS targets should be able to just link against minizip (whether that ends up in-tree or out) and it *should* just work (although I'm unable to test this). If i refactor zipped rom loading out of Core/, it could go into SDL/, as you suggested, but then iOS and Cocoa would be missing the feature. This could be solved by adding a new layer/module 
(could call it "Common/" or similar), but someone else would have to add support for that to the iOS/Cocoa code, since I don't have a mac.  So you have three choices: A) accept this as-is; B) have me move zipped rom loading into SDL/, dropping support for Apple devices, or, C) move it into a new "Common/" layer in-between SDL/ and Core/, and have someone who owns a Mac add support for that there.

(Sorry if this was already obvious when you asked me to move it to SDL, I'm new here ;)

- The Windows build environment, GnuWin32, looks like it was last updated in 2010; and it's version of zlib is from 2005 and doesn't contain minizip. Is there any interest in migrating to MSYS2/MinGW-w64/Cygwin? In that case, we could keep minizip out of tree, which would resolve the licensing issue. (I tried putting minizip in tree locally, and it didn't simplify the makefile at all, as it still depends on external zlib, which you have to test for anyway.)

I have a feeling it will be really easy to get SameBoy working under MSYS2 or one of the others, in fact I wouldn't be surprised if it compiles cleanly as-is, in which case we just need to update the build faq and nothing else.

brianmwaters:
fwiw, my vote is to create a "Common/" module, migrate to MSYS2/MinGW, and keep minizip out of tree.

I guess we still have to figure out the licensing thing in order to distribute binaries to Windows users? *groan*

sredna:
Windows has basic native zip support (enum and extract), just use `IShellFolder`. Should work as far back as XP, maybe even 2000.

TomTurbine:
Not exactly this exact topic but close enough to put here but.

While Zip is a common file, 7-zip seems to be increasingly used due to its higher compression ratio. Wouldn't it be a good idea to include this format with supporting other compressed archives?

Jack54guythecoder:
Are there any more problems that exist?

PR #536, opened 2023-03-30
Changed Makefile to work on recent MacOS without XCode

scanyard:
Currently, the Makefile fails if:

1) The user doesn't have XCode
2) The installed OS is a recent version of MacOS with a version higher than `10`

I've changed the `Makefile` to use `xcrun --show-sdk-path` instead. This has the added advantage of, if the tools aren't there, popping up a prompt on the OS for the user to install them (unlikely, if they already have `git` and `make`, but still possible).

Tested both the happy path of "SameBoy" still builds, as well as the error path `xcrun --show-sdk-path` returning a multiline error message, rather than the expected path. Make's string functions can be funky, but I've added an explanatory comment.

Your CONTRIBUTING file didn't say much about contributing to `Makefile`, so LMK if I can/need to change anything.


LIJI32:
Can't we just have `ifeq (,$(SYSROOT))` rather than `ifeq (,$(findstring /Library/Developer/CommandLineTools/SDKs,$(SYSROOT)))`? I expect `SYSROOT` to be an empty string if `xcrun --show-sdk-path` failed.

scanyard:
Unfortunately, `xcrun --show-sdk-path` returns a bunch of crap if there's no SDK installed. Let me see if I can make it behave; I should have time to update this in the next couple days.

Issue #548, opened 2023-06-07
Library access to raw APU channel volumes


nununoisy:
I'm currently working on a music visualizer program for the GameBoy in Rust, similar to my nsf-presenter project. I wanted to use SameBoy since it has good sound accuracy and is integration-friendly, so I wrapped the SameBoy core into a library crate and created Rust bindings for it with `bindgen`. I've verified that it works on Windows (after a lot of pain configuring `rust-cc` to use MSBuild+Clang), but I imagine that it would also work on Unix-y platforms without too much work.

I would like to visualize the channels by providing oscilloscope views of each channel as well as a piano roll (similar to Synthesia/other MIDI visualizers). To that end, it would be incredibly convenient to have access to the APU samples for each individual channel before mixing as well as a way to determine the fundamental frequency/pitch of the channel.

This should all be *technically* possible now if I use CGB mode w/ direct accesses to the IO registers:
- Channel volumes from `PCM12`/`PCM34` registers
- Frequencies from calculations on `NRx3` registers (not sure how the wave channel will work yet)

However, it would be far nicer to be able to access this data more directly, and have the ability to access it on anything that isn't a CGB. I'm not sure if it would be easier to add methods to directly query the APU (so one could sample at their leisure), add a new callback (so samples could be buffered during a `GB_run_frame()`), or something else entirely.

nununoisy:
I went ahead and implemented a hacky version of the setup using APU registers with a memory write callback for the `NRx1`/`NRx3`/`NRx4` registers and a couple of `GB_safe_read_memory()` calls for the `PCM12`/`PCM34` registers. It doesn't quite have all the features I need, unfortunately. However, it has shown me what an API could look like for most usecases like mine. Here's a hypothetical interface:

```c
typedef enum {
    GB_DUTY_CYCLE_12_5,
    GB_DUTY_CYCLE_25,
    GB_DUTY_CYCLE_50,
    GB_DUTY_CYCLE_75
} GB_duty_cycle_t;

typedef enum {
    GB_CHANNEL_PAN_NONE,
    GB_CHANNEL_PAN_L,
    GB_CHANNEL_PAN_R,
    GB_CHANNEL_PAN_L_R
} GB_channel_pan_t;

// Pre-mixer channel volume
uint8_t GB_get_channel_amplitude(GB_gameboy_t *gb, GB_channel_t channel);

// Value of the 11-bit period (NRx3/NRx4)
[...]

PR #549, opened 2023-06-14
Library access to raw APU channel volumes

nununoisy:
API implementation for the non-redundant methods laid out in #548.

This includes methods for fetching channel volume, pre-mix amplitude, and period, alongside the wave table with entry splitting as well as edge trigger detection. These are useful for visualizing APU data, e.g. for oscilloscope views of individual channels.

### DIFF
--- a/Core/apu.c
+++ b/Core/apu.c
@@ -1761,6 +1761,7 @@ bool GB_is_channel_muted(GB_gameboy_t *gb, GB_channel_t channel)
     return gb->apu_output.channel_muted[channel];
 }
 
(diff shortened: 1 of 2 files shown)


LIJI32:
Looks good, just two things:
1. Move the `edge_triggered` variables from `apu_t` to `apu_output_t` – these are temporary variables that don't affect emulation, so they shouldn't be in `apu_t` which is saved into save states.
2. Styling – the opening `{` of a function definition should be on its own line. (The rest of the `{`s are well-positioned)

LIJI32:
Merged, thanks!

PR #557, opened 2023-08-02
Add SIGUSR1 to reload to the SDL frontend

arkie:
Found this useful for quickly reloading after a change / compile - feel free to edit or close out if not wanted. Thanks!

### DIFF
--- a/SDL/main.c
+++ b/SDL/main.c
@@ -555,6 +555,11 @@ static void debugger_interrupt(int ignore)
     GB_debugger_break(&gb);
 }
 
+static void debugger_reset(int ignore)
+{
+    pending_command = GB_SDL_RESET_COMMAND;
+}
+
 static void gb_audio_callback(GB_gameboy_t *gb, GB_sample_t *sample)
 {    
     if (turbo_down) {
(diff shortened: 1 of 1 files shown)


arkie:
I used SIGHUP as it seems somewhat common for reload (e.g. I know [kanshi](https://git.sr.ht/~emersion/kanshi/tree/master/item/event-loop.c#L153) uses it, and a quick search found similar usage for HashiCorp's [Vault](https://support.hashicorp.com/hc/en-us/articles/5767318985107-Vault-SIGHUP-Behavior) and haproxy's [docker](https://hub.docker.com/_/haproxy))

Using `SIGUSR[12]` would work as well, was mostly just following that convention (and haven't yet run into a non-manual `SIGHUP` trigger issue heh)

LIJI32:
I assumed that was the reason, since SIGHUP was reused for these things historically since USR1 and USR2 are a relatively recent addition. Opening SameBoy SDL from the terminal and closing the terminal window triggers a SIGHUP, so the current code will trigger a restart rather than a termination in that case (None of the options are ideal to be fair, since the latter will quit without saving while the former is quite unexpected). I'd use SIGUSR1 for this, if you can modify the PR I'll merge it.

arkie:
Despite double-checking [Wikipedia](https://en.wikipedia.org/wiki/SIGHUP) re: history before sending the PR, I was still thinking "ah, of course killing the terminal will kill the process since it owns the session, so the behavior should still be intact..."

Well, at least it was a good reminder on `nohup` vs `setsid`. Thanks for the review!

PR #561, opened 2023-09-02
Force parsing order

vailkyte:


vailkyte:
MinGW randomly flakes out with me during a ternary. Just being safe here.

PR #562, opened 2023-09-03
Ternary safety

vailkyte:
Stop MinGW from breaking things!

### DIFF
--- a/Core/apu.c
+++ b/Core/apu.c
@@ -311,7 +311,7 @@ static void update_square_sample(GB_gameboy_t *gb, GB_channel_t index)
         return;
     }
 
-    uint8_t duty = gb->io_registers[index == GB_SQUARE_1? GB_IO_NR11 :GB_IO_NR21] >> 6;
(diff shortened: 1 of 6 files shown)


LIJI32:
Is MinGW legit compiling these wrong? Isn't this just GCC?

vailkyte:
Using stock g++-mingw-w64-i686-posix on ubuntu jammy - says gcc 10.

Something like this
`retro_debug_timer = strcmp(var.value, "On") == 0 ? 150 : 0;`

would always keep printing debug messages to RETRO_ENVIRONMENT_SET_MESSAGE (something I added). When I added the () guards around the strcmp test, then it correctly toggled on and off.

Then I started checking the sprite priority bug I was having (always was using X priority). Same problem and solution.

And noticed broken libretro joypad2, odd pixel glitches and sometimes weird sounds. These didn't happen with an earlier mingw build I had lying around - I think it was gcc 9.

So I thought it was worthwhile to notify upstream, and probably other libretro cores too...

vailkyte:
Think I goofed one of the () brackets. Have to recheck again.

PR #568, opened 2023-10-13
Warn user about SDL initialization failure

hitomi-nakayama:
On my system, I had issues with SameBoy detecting my controller, though sound and graphics were working.
It turns out that it was because `SDL_Init` was silently failing, as my system's SDL was compiled without support for haptics.
(My Gentoo Linux system's `libsdl2` was built without the `haptic` USE flag. After re-installing with the `haptic` flag enabled, my controller worked.)

This patch will now give the user this warning: `Couldn't initialize SDL: SDL not built with haptic (force feedback) support`.


hitomi-nakayama:
This may not be the best solution to the overall issue.

We could additionally warn the user that the lack of haptics support may lead to no controller functionality.
Alternatively, we can conditionally enable haptics support for SameBoy compilation, so that SDL initialization will not fail on these systems.

ISSOtm:
Perhaps it may be more useful to call [`SDL_InitSubSystem`](https://wiki.libsdl.org/SDL2/SDL_InitSubSystem) for each thing we want to init, and report errors for each. The increased granularity then would make the error messages more useful.

The questions then are:
- Would @LIJI32 accept that? It's a bit more code than there currently is.
- How should the errors be reported? Printed to the console, [`SDL_ShowSimpleMessageBox`](https://wiki.libsdl.org/SDL2/SDL_ShowSimpleMessageBox), both?

Issue #578, opened 2023-12-31
Update macOS sdk in Makefile


Buggem:
Error occured while building SameBoy:
```
Makefile:252: *** Could not find a macOS SDK.  Stop.
```
I went to line 252 in the Makefile and did some tinkering and realised that the Makefile program couldn't find the SDK because it specifically wanted version 10 and I only had 13, 13.3, 14, etc.
I had to modify the number on the Makefile manually, as I am on Sonama and I cannot install the older MacOSX10.sdk file. I modified the Makefile to use MacOSX13.sdk instead, and that worked.

Buggem:
See #579 for a potential fix.

PR #579, opened 2023-12-31
Fix bugs in Makefile for newer operating systems

Buggem:
Follow up to my issue [#578](https://github.com/LIJI32/SameBoy/issues/578#issue-2060974860). A fix for 2 bugs I was facing when I built SameBoy on macOS Sonama.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -245,8 +245,10 @@ CODESIGN := codesign -fs -
 else
 ifeq ($(PLATFORM),Darwin)
 SYSROOT := $(shell xcodebuild -sdk macosx -version Path 2> $(NULL))
+$(info "Password prompt to fix xcode-select bug.")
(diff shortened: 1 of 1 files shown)


Buggem:
Note: this patch is only intended to be a fix if worse comes to worse, as it requires a password. But as it is only for building, this might not matter as much.

LIJI32:
This isn't very future (or past) proof, as the user might only have the 12.x or 14.x SDKs, `grep "[0-9]\."` is probably a better solution to get the latest numbered SDK.

Buggem:
@LIJI32 updated as per code example

LIJI32:
I actually missed the `$(shell sudo xcode-select -s /Applications/Xcode.app/Contents/Developer)` line, why is this needed? It runs even if the previous line successfully found a Sysroot, and the following line will work regardless of the path reset.

PR #597, opened 2024-03-09
Fix -Wmissing-braces warnings

orbea:
I noticed these new warnings when backporting the strict-aliasing fix and they seem trivial to fix , with this I don't see any warnings with gcc or clang on my system.

### DIFF
--- a/Core/apu.c
+++ b/Core/apu.c
@@ -127,9 +127,9 @@ static void update_sample(GB_gameboy_t *gb, GB_channel_t index, int8_t value, un
         if (gb->io_registers[GB_IO_NR51] & (0x10 << index)) {
             left_volume = ((gb->io_registers[GB_IO_NR50] >> 4) & 7) + 1;
         }
-        GB_sample_t output = {0, 0};
+        GB_sample_t output = {{0, 0}};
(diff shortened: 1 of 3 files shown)


orbea:
@LIJI32 I would really appreciate if you could review this PR.

LIJI32:
Sorry, I haven't been available for work on SameBoy for a few weeks. I'm actually generally against keeping `-Wmissing-braces` enabled, as it's one of those warnings that don't ever detect actual bugs, and having to keep track of which structs require which amount of opening braces gets annoying.

Is there any particular reason you need/want these warnings fixed?

orbea:
@LIJI32 No worries, thanks for the reply.

I believe this basically boils down to correctness and being pedantic. I noticed these warnings because they were not present until recently and downstream projects didn't copy `-Wno-missing-braces` until now. My preference would be to use as few `-Wno-` flags as possible, but I will defer to your judgement.

LIJI32:
Fair. There are a few warnings that are normally enabled by `-Wall` that I tend to disable as they warn about "valid" C features I use a lot, such as `-Wint-in-bool-context` (handy shorthand, `-Wmissing-braces`) or `-Wmultichar` (Commonly used when handling file formats).

PR #603, opened 2024-03-22
Fix message always being logged when running ticks command

YayIguess:
"Tick count reset." is currently logged even if the user supplied the 'keep' argument to the ticks command in the debugger.


LIJI32:
Good catch, thanks!

PR #607, opened 2024-04-13
Fix potential compiler warning on questionable code path

Morilli:
This line displayed a compiler warning to me: 
```
.../Core/gb.c:1708:27: warning: arithmetic on a null pointer treated as a cast from integer to pointer is a GNU extension [-Wgnu-null-pointer-arithmetic]
 1708 |     memset(gb, 0, (size_t)GB_GET_SECTION(NULL, unsaved));
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
.../Core/save_state.h:24:59: note: expanded from macro 'GB_GET_SECTION'
   24 | #define GB_GET_SECTION(gb, name) (void *)((uint8_t *)(gb) + GB_SECTION_OFFSET(name))
      |                                           ~~~~~~~~~~~~~~~ ^
```

The warning probably only surfaces with specific compiler/warning combinations used, but this line of code seems questionable anyways considering what it's actually trying to do, so I've simplified it and fixes the compiler warning at the same time.

### DIFF
--- a/Core/gb.c
+++ b/Core/gb.c
@@ -1705,7 +1705,7 @@ static void GB_reset_internal(GB_gameboy_t *gb, bool quick)
     GB_update_clock_rate(gb);
     uint8_t rtc_section[GB_SECTION_SIZE(rtc)];
     memcpy(rtc_section, GB_GET_SECTION(gb, rtc), sizeof(rtc_section));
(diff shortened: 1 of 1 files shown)


LIJI32:
I probably wrote this before I added the `GB_SECTION_OFFSET` macro. Thanks!

PR #631, opened 2024-06-02
call SDL_Quit() on exit to prevent a crash on app exit with linux/wayland

laamaa:
On Bluefin DX Linux (a Fedora Silverblue spinoff) closing the app window brings down the whole Gnome session. Ensuring that SDL_Quit() (https://wiki.libsdl.org/SDL2/SDL_Quit) is called on exit seems to prevent this. The problem happened with both Flatpak and self-compiled versions

### DIFF
--- a/SDL/main.c
+++ b/SDL/main.c
@@ -1052,7 +1052,9 @@ int main(int argc, char **argv)
 #endif
 
     SDL_Init(SDL_INIT_EVERYTHING & ~SDL_INIT_AUDIO);
-    atexit(SDL_Quit);
+    // This is, essentially, best-effort.
+    // This function will not be called if the process is terminated in any way, anyhow.
+    (void)atexit(SDL_Quit);
 
     if ((console_supported = CON_start(completer))) {
         CON_set_repeat_empty(true);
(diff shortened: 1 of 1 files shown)


SDL/main.c
```diff
@@ -1052,6 +1052,8 @@ int main(int argc, char **argv)
 #endif
 
     SDL_Init(SDL_INIT_EVERYTHING & ~SDL_INIT_AUDIO);
+    atexit(SDL_Quit);
```
ISSOtm, line 1055:
```suggestion
    // This is, essentially, best-effort.
    // This function will not be called if the process is terminated in any way, anyhow.
    (void)atexit(SDL_Quit);
```

ISSOtm (changes requested):
This works for me under Linux+Wayland (sway), but so does `master` FWIW. `sway` does not log any errors.

I suggest you also report a bug to the GNOME developers&mdash;I do not think that the *entire session* should be *able* to be brought down by a single misbehaving program.

Though this workaround is still worth merging, IMO. Albeit I made some context-adding suggestions, subject to your convenience.

laamaa:
Good idea, I added the suggestions to the commit. The issue doesn't seem to happen on vanilla Silverblue/Gnome (without proprietary NVIDIA drivers installed, unlike Bluefin), but I remember bumping into the same problem at some point in my own SDL based project, so it's probably a good idea to have it there just in case...

ISSOtm (approved):
Seems fair to me, then :)

PR #638, opened 2024-06-20
Update bootrom code with the latest hardware.inc

Rangi42:
Bootrom code is updated to use new hardware.inc constants, as well as for consistent formatting.

The ROMs are unchanged:

```
$ sha1sum build/bin/BootROMs/*.bin
4ed1d95c9c682a78698f235ce9e9bb3b62bccf9f  build/bin/BootROMs/agb_boot.bin
822728fbe1d637528c57659ad86e0bf5da666677  build/bin/BootROMs/cgb0_boot.bin
113903775a9d34b798c2f8076672da6626815a91  build/bin/BootROMs/cgb_boot.bin
1db57a1e8b6e4096f811587f9eab0c6675fd9755  build/bin/BootROMs/dmg_boot.bin
9817bdae9335a3accb584f77c165dc5726555d1f  build/bin/BootROMs/mgb_boot.bin
f282b3aaf98f8423dab7d77f1aa0192be630f2fb  build/bin/BootROMs/sgb2_boot.bin
369e6eb5e0c975eaa52a4a3f6ee07b2a3c3c16de  build/bin/BootROMs/sgb_boot.bin
```

These should still build with RGBDS 0.5.0 through 0.8.0 (master). Compatibility notes:

- 0.8.0 will deprecate `-i`, requiring `-I`, so we just use the long `--include` option.
- Using `halt` without an implicit `nop` after it requires version-specific flags. RGBASM 0.5.0 and 0.6.0 needed `-h` to disable it; 0.7.0 does so by default. (0.7.0 lets you opt in with `-H`, but 0.8.0 will remove that entirely.)


BootROMs/cgb_boot.asm
```diff
@@ -1,27 +1,27 @@
...
 
 ; Clear memory VRAM
-    call ClearMemoryPage8000
+    call ClearMemoryVRAM
 
 ; Clear OAM
-    ld h, $fe
-    ld c, $a0
+    ld h, HIGH(_OAMRAM)
+    ld c, $A0
```
ISSOtm, line 15: This can be
```suggestion
    ld c, 4 * OAM_COUNT
```
or even
```suggestion
    ld c, sizeof_OAM_ATTRS * OAM_COUNT
```

BootROMs/cgb_boot.asm
```diff
@@ -709,13 +730,13 @@ LoadTileset:
     ld c, a
     jr nc, .shift_left
     srl a
-    db $fe ; eat the add a with cp d8
+    db $FE ; eat the add a with `cp d8`
```
ISSOtm, line 733:
```suggestion
    db $FE ; eat the `add a` with `cp d8`
```

BootROMs/cgb_boot.asm
```diff
@@ -709,13 +730,13 @@ LoadTileset:
...
 .shift_left
     add a
     sla b
     jr c, .go_and
     or c
-    db $fe ; eat the and c with cp d8
+    db $FE ; eat the and c with `cp d8`
```
ISSOtm, line 739:
```suggestion
    db $FE ; eat the `and c` with `cp d8`
```

ISSOtm (changes requested):
Nice thinking for `--include`!

ISSOtm (changes requested):

LIJI32:
Great work! I've been wanting to eventually clean SameBoot up and use hardware.inc more "fully", but I never had the chance, and you probably did a far better job than I would. I do have a few comments though:
1. The Makefile has a few unnecessary whitespace change, can you rebase them away so they don't screw `git blame` up?
2. I'd have a Make variable for RGBASM flags instead of having them inline calculated on build. It looks a bit cleaner and is slightly faster.
3. `ld hl, (BOOTUP_A_MGB << 8) | $B0` and co. could use the `lb` macro.
4. `CgbFlag`, `SgbFlag`, `RomSize` and `RamSize` should be `CGBFlag`, `SGBFlag`, `ROMSize` and `RAMSize`, respectively.

ISSOtm (approved):

PR #665, opened 2024-10-21
ios/tvos: properly set min supported version

warmenhoven:


LibretroAdmin:
HI, we'd appreciate it very much if this could be merged soon. We need it on the CI end

PR #681, opened 2025-01-23
correct mimetype location to follow the shared mime info specification

Jan200101:
according to the shared-mime-info specification it should be installed into the packages folder:

> Each application that wishes to contribute to the MIME database will install a single XML file, named after the application, into one of the three <MIME>/packages/ directories (depending on where the user requested the application be installed). After installing, uninstalling or modifying this file, the application MUST run the update-mime-database command, which is provided by the freedesktop.org shared database[[SharedMIME](https://specifications.freedesktop.org/shared-mime-info-spec/latest/bi01.html#id-1.5.6)].

[Unified system | Shared MIME-info Database](https://specifications.freedesktop.org/shared-mime-info-spec/latest/ar01s02.html#s2_layout)

I saw that this change was made for FreeBSD but they seem to support the shared mime database
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -783,7 +783,7 @@ install: $(BIN)/XdgThumbnailer/sameboy-thumbnailer sdl $(shell find FreeDesktop)
 	install -d $(DESTDIR)$(DATA_DIR)/BootROMs
 	install -d $(DESTDIR)$(PREFIX)/bin
 	install -d $(DESTDIR)$(PREFIX)/share/thumbnailers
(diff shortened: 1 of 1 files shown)


LIJI32:
It's probably wise to uninstall the copy outside of `packages/` that might have been there from a previous installation in the `install` target, otherwise you'd have two competing definitions.

Jan200101:
I'm not so sure...
the codepath is only accessed when DESTDIR is set which I don't think a regular user would do, only system packages appear to be impacted by this changes

PR #695, opened 2025-04-04
Add missing break statements in mbc7 writes

nik012003:
Small fix since it would be highly unlikely to be a problem in a game.


LIJI32:
Good catch, thanks!

PR #697, opened 2025-04-05
Don’t look for libdl on OpenBSD

bentley:
Like Haiku, OpenBSD doesn’t provide libdl.


LIJI32:
I wonder if the "correct" solution would be to only do this on Linux, because `libdl` honestly sounds very glibc specific at this point.

bentley:
Probably… Even environments that provide libdl seem to mostly do so for compatibility with glibc, and don’t actually require it be used.

PR #700, opened 2025-04-30
SPM support for Core

szotp:
Added basic support for Swift Package Manager. SPM can handle C libraries, and Swift supports binding to them, which means someone else can implement a wrapper on top of it. 

I had to make some adjustments in headers:
- Unnecessary circular dependency of `display.h` on `gb.h`
- `apu.h` using `GB_ENUM` without including `save_state.h`
- `save_state.h` using `GB_gameboy_t` without including `model.h`

### DIFF
--- a/Core/apu.h
+++ b/Core/apu.h
@@ -4,6 +4,7 @@
 #include <stddef.h>
 #include <stdio.h>
 #include "defs.h"
+#include "save_state.h"
 
 #define GB_BAND_LIMITED_WIDTH 16
 #define GB_BAND_LIMITED_PHASES 512
--- a/Core/display.h
+++ b/Core/display.h
@@ -1,6 +1,5 @@
 #pragma once
 
-#include "gb.h"
 #include <stdbool.h>
 #include <stdint.h>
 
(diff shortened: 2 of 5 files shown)


szotp:
I tried to run `https://pub.dev/packages/ffigen` and it found further errors with missing includes.

Pushed fixes for those. Maybe it would make sense to add 
```swift
#include <stdbool.h>
#include <stdint.h>
#include <stddef.h>
```

into defs.h and only include one file in other places?

LIJI32:
I have several issues and comments:

1. Header cleanup is good, can be merged pretty much as is.
2. I'm rather worried about adding SPM support to the mainstream repo. I don't use Swift and I don't know how SPM works, so I couldn't possibly provide official support to it. Could this be done in a different repo with SameBoy being a subrepo?
3. The "ideal" way of using SameBoy as a library is using `make lib` (potentially with `CONF=release` and potentially extra CFLAGS/LDFLAGS/SYSROOT for cross-compilation if needed) and using the generated a/o/dylib file together with the post-processed headers. This makes sure SameBoy is built with LTO and the optimal optimization flags, but I worry the most about having `GB_INTERNAL` defined globally like you do here – if I understand SPM correctly, this would expose internal APIs to the Swift side, which is quite problematic.

PR #705, opened 2025-05-27
Add support for Vim-style hjkl D-pad keys

guitaripod:
This enables Vim navigation for D-pad movement in-game, for the macOS and SDL versions of the emulator.

My mind is mapped to Vim, so this feels much more natural than arrow keys, which remain functional as they are.


marcusziade:
Honestly, I think what happened was that I created the first PR, then played a game, realized that hjkl didn't work, and just made this PR without thinking it through too much.

Issue #698, opened 2025-04-06
ISO C++ forbids declaration of 'typeof' with no type


carmiker:
After updating bsnes-jg to use SameBoy 1.0.1, I am now receiving this error at build time:
```
In file included from /home/rupert/src/bsnes/src/coprocessor/icd.cpp:31:
/home/rupert/src/bsnes/deps/gb/gb.h:235:14: error: ISO C++ forbids declaration of 'typeof' with no type [-fpermissive]
  235 | static const typeof(GB_IO_PGB) __attribute__((deprecated("Use GB_IO_PGB instead"))) GB_IO_UNKNOWN5 = GB_IO_PGB;
````
bsnes-jg is ISO C++11, and `typeof` is a GNU extension. I would like to avoid building with gnu++11 if possible. For the moment, I've just removed the line in question from `gb.h`. Alternatively, I have wrapped it in `#ifndef __cplusplus`, but I am not sure if that would be acceptable upstream. Perhaps there is a better way to manage this situation.

LIJI32:
Pretty sure `#define typeof decltype` inside `if defined(__cplusplus) && defined (__STRICT_ANSI__)` will do the trick?

carmiker:
Yes, this works. In my case, adding `#define typeof decltype`  before including `gb.h` is an even simpler solution.

LIJI32:
I'm leaving this open because I want to properly do it in `gb.h` itself so it's more straightforward to other affected users of the header.

PR #710, opened 2025-07-05
fix typeof usage in gb.h

Morilli:
`typeof` is not part of the C standard until C23, so `__typeof__` should be used instead. `defs.h` has a `#define typeof __typeof__` line, but it's guarded behind `GB_INTERNAL` which may not be defined when #include'ing `gb.h` from some external place.

I ran into a compiler error with this line and spent too much time debugging and understanding the code and compiler behavior before realizing what the problem was. Alternatives to this would be using the `c23` standard in the project instead, just typing `int` instead of using `typeof` or not guarding the define in defs.h behind `GB_INTERNAL`. I'm not making an issue for this because it's very easy to fix, just super annoying to stumble over (the compiler error message is NOT helpful).

### DIFF
--- a/Core/gb.h
+++ b/Core/gb.h
@@ -235,7 +235,7 @@ enum {
     GB_IO_PCM34      = 0x77, // Channels 3 and 4 amplitudes
 };
 
-static const typeof(GB_IO_PGB) __attribute__((deprecated("Use GB_IO_PGB instead"))) GB_IO_UNKNOWN5 = GB_IO_PGB;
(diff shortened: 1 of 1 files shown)


LIJI32:
See also this issue: https://github.com/LIJI32/SameBoy/issues/698

LIJI32:
It might be already defined to `__typeof__`, `decltype`, or `__typeof` (which is also valid), so I do that to avoid erroring out over a duplicate `#define`.

LIJI32:
Also, what compiler and flags you're using? `__STRICT_ANSI__` is normally defined when using `-std=cXX`.

PR #722, opened 2025-09-02
Build boot ROMs with RGBDS 0.9.4 or greater

Rangi42:
RGBASM 0.9.4 introduced `'c'`haracter literals as a first step toward separating string and numeric expressions. RGBASM 1.0.0 will deprecate using single-character strings as numbers, which causes `make bootroms` to print deprecation warnings.

This does mean that the oldest supported RGBDS version will now be 0.9.4. However, that version is available from gbdev's releases, macOS `brew`, Arch `pacman`, and Gentoo `portage`, so requiring it will hopefully not be a problem.

If SameBoy BootROMs still want to support older RGBDS, I can change this PR to use ASCII codes `$30` and `$31` instead of `'0'` and `'1'`, and retain the Makefile support for older flags.

image


LIJI32:
Dropping support for older RGBDS version is problematic because obtaining an up-to-date version of RGBDS is actually very annoying on older OS versions and certain package managers. A notable example of this is Brew on any macOS version that isn't one of the latest 3 majors. On these versions, Brew falls back to source-based package managing, meaning it will build every single RGBDS dependency, and since the Brew package includes the Rust-based tools as well, it will take *ages* and a lot of disk space to build, as it will build LLVM and the Rust compiler as well.

Fortunately, with official non-Brew binary releases of RGBDS, this isn't a problem on macOS (And SameBoy explicitly recommend using this method of obtaining RGBDS), but this is still potentially a problem on some Linux distros, and an inconvenience for people who only use RGBDS for building SameBoy. So I'd rather keep RGBDS 0.6 supported, since it isn't too costly to support anyway.

That said, updating the CI to use 1.0 is probably a good idea.

Rangi42:
> Dropping support for older RGBDS version is problematic ...
> 
> That said, updating the CI to use 1.0 is probably a good idea.

No problem, updated to avoid the new 0.9.4 feature of `'c'`haracter literals.

PR #732, opened 2025-11-12
Removal of cheats should be based on cheat_count, not gb_cheats

LumioseSil:
This is because we call free based on cheat_count, not if cheats is not null.


LIJI32:
These conditions are equivalent (`gb->cheat_count` is non-zero IFF `gb->cheats` is non-NULL), but it mutes Clang analyzer's false positive.

PR #760, opened 2026-06-15
Core: Fix object_low_line_address masking in sanitize_state

alexemorris:
In sanitize_state, gb->object_low_line_address was incorrectly masked with gb->vram_size & ~1. If vram_size is 16KB (0x4000), this masks out all bits except bit 14, causing it to lose address bits and trigger out-of-bounds reads. Changed to mask with (gb->vram_size - 1) & ~1.


LIJI32:
Good catch, thanks!

